PR #21, opened 2025-07-03
Apg 1060/caselist UI

niaevans9:
Setting up the basic UI components needed for the caselist Page -
Adding scripts to generate Types for the gov frontend components
Tabs which call out to the API endpoints to receive open and closed referrals
Filters which are populated with the cohort and status values (will also generate selected filter pane and correct url params when the button is activated)
Search by Name or CRN filter
Pagination component included ready for when the real API endpoint is completed
Screenshot 2025-07-03 at 14 29 00


pfcairns (dismissed):

thomaswilsonxyz (dismissed):
left a few comments, I think we can merge through then address these in a follow-up PR, let's keep to many smaller changes

PR #33, opened 2025-07-15
APG-1056: Generate types from API and implement pagination for caselist

Jack-Dempster:
This adds pagination to the caselist view
image

It also integrates with the API endpoint to retrieve referrals from the `Referral case list item view` and so we can now click into the referral details page.


pfcairns (approved):
LGTM

niaevans9 (approved):
lgtm

PR #58, opened 2025-07-30
APG-1167: Filter ethnicity field if it's not present

Jack-Dempster:
We have discovered that ethnicity can be null as part of the nDelius personal details response. This PR ensures that we hide that field if `ethnicity` is not present.

image


niaevans9 (approved):
lgtm

PR #66, opened 2025-08-11
Apg 1126/offence history tab UI

niaevans9:


server/@types/manageAndDeliverApi/imported/index.d.ts
```diff
@@ -409,6 +433,42 @@ export interface components {
        */
       dateRetrieved: string
     }
+    /** @description Details of an offence committed by an offender */
```
thomaswilsonxyz, line 436: Is this `an` offence, or is it the offence associated with the Referral ?

server/@types/manageAndDeliverApi/imported/index.d.ts
```diff
@@ -513,12 +602,48 @@ export interface components {
...
       /** @enum {string} */
       overallRelationshipDomainLevel?: 'HIGH_NEED' | 'MEDIUM_NEED' | 'LOW_NEED'
       individualRelationshipScores: components['schemas']['IndividualRelationshipScores']
     }
+    RiskScore: {
+      /**
+       * @description classification associated with PNI Eg. HIGH_RISK, MEDIUM_RISK, LOW_RISK
+       * @example High Risk
+       */
+      classification: string
```
thomaswilsonxyz, line 620: If we know the possible values (as above) _could_ we have this as a type literal ?

```ts
classification: 'HIGH_RISK' | 'MEDIUM_RISK' | 'LOW_RISK'
```

server/@types/manageAndDeliverApi/imported/index.d.ts
```diff
@@ -513,12 +602,48 @@ export interface components {
...
+      /** @example 2 */
+      IndividualRiskScores: components['schemas']['IndividualRiskScores']
+    }
+    Sara: {
+      /**
+       * @description The highest of what is being returned based on saraRiskOfViolenceTowardsPartner and saraRiskOfViolenceTowardsOthers
+       * @example LOW
+       * @enum {string}
```
thomaswilsonxyz, line 628: duplicate info

server/@types/manageAndDeliverApi/imported/index.d.ts
```diff
@@ -853,6 +978,65 @@ export interface operations {
...
+        /** @description The id (UUID) of a referral */
+        id: string
+      }
+      cookie?: never
+    }
+    requestBody?: never
+    responses: {
+      /** @description Information about the offence history */
+      200: {
+        headers: {
```
thomaswilsonxyz, line 995: you could also have `Record<string, unknown>`

server/@types/manageAndDeliverApi/imported/index.d.ts
```diff
@@ -853,6 +978,65 @@ export interface operations {
...
+        }
+      }
+      /** @description Bad Request */
+      400: {
+        headers: {
+          [name: string]: unknown
+        }
+        content: {
+          '*/*': components['schemas']['ErrorResponse']
+        }
```
thomaswilsonxyz, line 1009: given the duplicate of this little object, might be worth extracting somewhere and re-using

server/referralDetails/offenceHistoryPresenter.ts
```diff
@@ -1,12 +1,23 @@
...
 import ReferralDetailsPresenter from './referralDetailsPresenter'
+import DateUtils from '../utils/dateUtils'
 
 export default class OffenceHistoryPresenter extends ReferralDetailsPresenter {
   constructor(
     readonly details: ReferralDetails,
     readonly subNavValue: string,
     readonly id: string,
+    readonly offenceHistory: {
+      mainOffence: { offence: string; offenceCode: string; category: string; offenceDate: string; categoryCode: string }
```
thomaswilsonxyz, line 12: formatting seems weird, we have new-lines below, but single line here

server/referralDetails/referralDetailsController.test.ts
```diff
@@ -63,13 +64,15 @@ describe('referral-details', () => {
...
+      const offenceHistory: OffenceHistory = offenceHistoryFactory.build()
+      accreditedProgrammesManageAndDeliverService.getOffenceHistory.mockResolvedValue(offenceHistory)
       return request(app)
         .get(`/referral-details/${randomUUID()}/offence-history`)
         .expect(200)
         .expect(res => {
           expect(res.text).toContain(referralDetails.crn)
           expect(res.text).toContain(referralDetails.personName)
-          expect(res.text).toContain('offence-history')
+          expect(res.text).toContain('Absconding from lawful custody')
```
thomaswilsonxyz, line 75: I don't like that this string (`Absconding...`) relies on a default value set in a factory, could we make sure to explicitly set this value in the test file itself.

in my experience, over time, having realistic data set as defaults in factories can make debugging harder b/c you're not _quite_ sure where a value is being set

personally I'll always default to obvious-default values (e.g. `"the factory default text"`) but I know that's not our pattern here.

At very least our test should be showing us that things that come back from the (mocked) API are making it through the pipeline to the `res` w/o being modified, and I can't assert (as a human developer) _just_ by looking at this file, I'd have to navigate elsewhere, and that's a point of friction we can remove.

server/services/accreditedProgrammesManageAndDeliverService.ts
```diff
@@ -110,4 +111,12 @@ export default class AccreditedProgrammesManageAndDeliverService {
       data: updateAvailabilityParams,
     })) as Availability
   }
+
+  async getOffenceHistory(username: Express.User['username'], referralId: string): Promise<OffenceHistory> {
+    const restClient = await this.createRestClientFromUsername(username)
+    return (await restClient.get({
+      path: `/referral-details/${referralId}/offence-history`,
+      headers: { Accept: 'application/json' },
+    })) as OffenceHistory
```
thomaswilsonxyz, line 120: more of an implementation question than anything, but will this handle the feeding through of the `4xx` responses you specify in `server/@types/manageAndDeliverApi/imported/index.d.ts` ?

those responses wouldn't be `OffenceHistory`, right ? they'd be one of the error responses

it might be worth having at least one unhappy path test (I'm not sure where, maybe the Controller) to show what happens if e.g. a `401` comes back

niaevans9: Currently the error handler would catch a 401 and sign out the person - not sure if this is the expected behaviour but its how the error handler behaves out of the box - Ive not seen any tickets around handling these kind of errors - I think in F&R we had specific pages
   if (error.status === 401 || error.status === 403) {
      logger.info('Logging user out')
      return res.redirect('/sign-out')
    }

server/referralDetails/referralDetailsController.test.ts
niaevans9: Test Updated

PR #116, opened 2025-09-09
Apg 1169/show location preferences

niaevans9:
Initial PR to add in location preferences UI and the link to access journey to select preferences
PR to follow to call API and populate data
Screenshot 2025-09-09 at 10 37 55


thomaswilsonxyz (approved):
leaving a few comments, nothing drastic, but just asking questions and checking decisions :)

PR #117, opened 2025-09-09
add risks and alerts sub nav

rgforsyth:
Adds content to the risks & alerts sub nav on the risks and alerts page.
Contains a few components. Custom, from risk-ui-elements and some taken and modified from custody.

image


server/testutils/factories/factoryHelpers.ts
```diff
@@ -0,0 +1,42 @@
+import { faker } from '@faker-js/faker/locale/en_GB'
```
rgforsyth, line 1: Taken from custody

PR #122, opened 2025-09-15
APG-1169: Show previously set delivery location preferences

Jack-Dempster:
This PR integrates with the new GET delivery location preferences endpoint so we can display previously set delivery location preferences for a referral. 

Screens below show both when we have no preferences set and when a referral does have preferences set
image

image

### DIFF
--- a/server/referralDetails/locationPresenter.ts
+++ b/server/referralDetails/locationPresenter.ts
@@ -16,22 +16,20 @@ export default class LocationPresenter extends ReferralDetailsPresenter {
     const { lastUpdatedAt, lastUpdatedBy, preferredDeliveryLocations, cannotAttendLocations } =
       this.deliveryLocationPreferences
 
-    const summaryItems: SummaryListItem[] = [
+    let summaryItems: SummaryListItem[] = [
       {
         key: 'Preferred programme delivery locations',
         lines: preferredDeliveryLocations.length !== 0 ? preferredDeliveryLocations : ['No information added'],
       },
       {
-        key: 'Locations the personal cannot attend',
-        lines: [cannotAttendLocations || 'No information added'],
+        key: 'Locations the persona cannot attend',
+        lines: [cannotAttendLocations ?? 'No information added'],
       },
     ]
 
     // Adds to the start of the arry if fields are not null
     if (lastUpdatedAt && lastUpdatedBy) {
-      summaryItems.unshift({
-        lines: [`Last updated ${lastUpdatedAt} by ${lastUpdatedBy}`],
-      })
+      summaryItems = [{ lines: [`Last updated ${lastUpdatedAt} by ${lastUpdatedBy}`] }, ...summaryItems]
     }
 
     return {
--- a/server/referralDetails/referralDetailsController.test.ts
+++ b/server/referralDetails/referralDetailsController.test.ts
@@ -141,6 +141,32 @@ describe('referral-details', () => {
           expect(res.text).toContain('Location')
         })
     })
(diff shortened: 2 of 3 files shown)


server/referralDetails/locationPresenter.ts
```diff
@@ -1,33 +1,43 @@
...
+      this.deliveryLocationPreferences
+
+    const summaryItems: SummaryListItem[] = [
+      {
+        key: 'Preferred programme delivery locations',
+        lines: preferredDeliveryLocations.length !== 0 ? preferredDeliveryLocations : ['No information added'],
+      },
+      {
+        key: 'Locations the personal cannot attend',
+        lines: [cannotAttendLocations || 'No information added'],
```
thomaswilsonxyz, line 26: can we use `??` not `||` here, it's a bit more flexible

server/referralDetails/locationPresenter.ts
```diff
@@ -1,33 +1,43 @@
...
+      },
+    ]
+
+    // Adds to the start of the arry if fields are not null
+    if (lastUpdatedAt && lastUpdatedBy) {
+      summaryItems.unshift({
```
thomaswilsonxyz, line 32: i hate mutative js array methods, b/c their return values and side-effects are so non-obvious from their signatures

```ts
summaryItems = [{lines: [`Last updated ${lastUpdatedAt} by ${lastUpdatedBy}`]}, ...summaryItems]
```

server/referralDetails/locationView.ts
```diff
@@ -1,12 +1,12 @@
...
 import ViewUtils from '../utils/viewUtils'
+import LocationPresenter from './locationPresenter'
 
 export default class LocationView {
   constructor(private readonly presenter: LocationPresenter) {}
 
   get getPreferredLocationsAsSummaryListArgs(): SummaryListArgs {
-    const summary = this.presenter.PreferredLocationsSummary()
+    const summary = this.presenter.preferredLocationsSummary()
     const addLocationPreferenceHref = `/referral/${this.presenter.referralId}/add-location-preferences`
```
thomaswilsonxyz, line 10: genuine curiosity: do we have a list of routes for the web app stored anywhere?

server/referralDetails/referralDetailsController.test.ts
```diff
@@ -118,9 +119,19 @@ describe('referral-details', () => {
   })
 
   describe(`GET /referral-details/:id/location`, () => {
-    it('loads the referral details page with locations sub-nav', async () => {
+    it('loads the referral details page with locations sub-nav with no existing details', async () => {
```
thomaswilsonxyz, line 122: do we need a sibling test which _does_ have the Delivery Location Preferences ?

thomaswilsonxyz (commented):
just left a few comments

assets/scss/overrides/_local.scss
```diff
@@ -17,7 +17,13 @@
 }
 
 .govuk-summary-card__content {
-  .no-first-key .govuk-summary-list__row:first-child .govuk-summary-list__key {
-    display: none;
+  .no-first-key .govuk-summary-list__row:first-child {
+    display: table-caption;
```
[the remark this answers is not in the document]
Jack-Dempster: @niaevans9 might be better to answer this one as I had to bring in the cavalry for this. But it's specifically for this box 
image

as we are using the summary component but need a row above that does not have the same styling as normal

thomaswilsonxyz (approved):

PR #128, opened 2025-09-17
Apg 1231/manually change cohort

anandelangovan1:


server/cohort/changeCohortPresenter.test.ts
```diff
@@ -0,0 +1,41 @@
...
+      },
+    })
+  })
+})
+
+/* eslint-disable import/prefer-default-export */
+export enum CohortEnum {
+  SEXUAL_OFFENCE = 'SEXUAL_OFFENCE',
+  GENERAL_OFFENCE = 'GENERAL_OFFENCE',
+}
```
Jack-Dempster, line 41: is this so we are explicitly setting in test? Just wondering why we aren't using the imported enum? Also shouldn't export this if it's not used elsewhere

server/cohort/changeCohortView.ts
```diff
@@ -0,0 +1,61 @@
...
+          classes: 'govuk-fieldset__legend--m',
+        },
+      },
+      items: [
+        {
+          value: 'SEXUAL_OFFENCE',
+          text: 'Sexual offence',
+        },
+        {
+          value: 'GENERAL_OFFENCE',
```
Jack-Dempster, line 53: can we not use the cohort enum for the values here?

server/risksAndNeeds/risksAndNeedsPresenter.ts
```diff
@@ -41,6 +41,12 @@ export default class RisksAndNeedsPresenter extends ReferralLayoutPresenter {
           classes: 'govuk-button--secondary',
           href: '/pdu/open-referrals',
         },
+
+        {
+          text: 'Update cohort',
+          classes: 'govuk-button--secondary',
+          href: `/referral/${this.referralId}/change-cohort`,
+        },
```
Jack-Dempster, line 49: since you've added the value to the `ReferralLayoutPresenter` I'm not sure we need this `getSubHeaderAtgs()` function at all anymore?

server/cohort/changeCohortPresenter.test.ts
anandelangovan1: I wanted to use the types which were available, but lint was complaining and it. 

Options were using an export or disabling the lint rule

PR #133, opened 2025-09-22
APG-1182: Update LDC status for a person

Jack-Dempster:
This PR adds the ability for a user to update the ldc status of a referral. Shown below are the screens that have been implemented.

image
image

## KNOWN ISSUES
- The success alert box when updating ldc status will only appear when updated on the personal details section page, this is an existing issue in the cohort journey as well and requires changes in many files which would muddy this PR.
- Query params are being duplicated if updating status multiple times in a row, this also exists in the cohort journey and for same reason above will address in another pr,.

### DIFF
--- a/server/ldc/updateLdcView.ts
+++ b/server/ldc/updateLdcView.ts
@@ -1,4 +1,4 @@
-import { InsetTextArgs } from '../utils/govukFrontendTypes'
+import { InsetTextArgs, RadiosArgs } from '../utils/govukFrontendTypes'
 import UpdateLdcPresenter from './updateLdcPresenter'
 
 export default class UpdateLdcView {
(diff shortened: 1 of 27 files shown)


server/ldc/updateLdcView.ts
```diff
@@ -0,0 +1,59 @@
...
+export default class UpdateLdcView {
+  constructor(private readonly presenter: UpdateLdcPresenter) {}
+
+  get renderArgs(): [string, Record<string, unknown>] {
+    return [
+      'referralDetails/ldc/updateLdc',
+      {
+        presenter: this.presenter,
+        currentLdcStatusText: this.currentLdcStatusText,
+        radioArgs: this.radioArgs.bind(this),
```
niaevans9, line 13: Unless we are passing values in from the njk file into the radioArgs function we dont need to .bind(this)

niaevans9 (approved):
lgtm

PR #143, opened 2025-09-29
Manage LDC and cohort messages from the parent presenter

niaevans9:
LDC and Cohort Success messages are now managed by the parent presenter
LDC and Cohort update values still need to be passed through to the parent presenter
Removed redundant passing of Referral IDs where we are also passing the referral object to reduce number of params being passed around
Redirect after successful LDC or Cohort update back to the page the user was originally on instead of the personal details page every time


Jack-Dempster (approved):
Looks great and way cleaner now

PR #453, opened 2026-04-15
Edit a groups PDU and location

niaevans9:
Add the edit group journeys for PDU and Locations

image
image
image
image


server/createGroup/editGroupController.ts
```diff
@@ -166,4 +170,109 @@ export default class EditGroupController extends BaseController {
     const view = new RescheduleSessionsView(presenter)
     return this.renderPage(res, view)
   }
+
+  async editGroupPdu(req: Request, res: Response): Promise<void> {
```
rgforsyth, line 174: Do we need to clear the session after the update has happened? Maybe not

niaevans9: When we redirect to the group details page after update it clears the session object
req.session.createGroupFormData = {}

server/createGroup/editGroupController.test.ts
```diff
@@ -379,4 +379,241 @@ describe('Edit Group Controller', () => {
...
+      return request(app)
+        .post(`/group/${groupId}/edit-group-probation-delivery-unit`)
+        .type('form')
+        .send({ 'create-group-pdu': JSON.stringify({ code: 'PDU-NW', name: 'North West' }) })
+        .expect(302)
+        .expect(res => {
+          expect(res.text).toContain(`Redirecting to /group/${groupId}/edit-group-delivery-location`)
+        })
+    })
+  })
```
raby, line 483: Could we add a test for the edit location page when originPage is not present in session? That would pin down the expected Back link behaviour and protect against the broken-navigation case if someone lands on this route without first coming from group details.

it('falls back to group details when originPage is missing from session', async () => {
  const sessionData: Partial<SessionData> = {
    createGroupFormData: {
      groupCode: 'TEST123',
      pduCode: 'PDU-NE',
      pduName: 'North East',
      deliveryLocationCode: 'LOC-1',
      deliveryLocationName: 'HMP Leeds',
    },
  }

  app = TestUtils.createTestAppWithSession(sessionData, { accreditedProgrammesManageAndDeliverService })

  accreditedProgrammesManageAndDeliverService.getOfficeLocationsForPdu.mockResolvedValue([
    { code: 'LOC-1', description: 'HMP Leeds' },
  ])

  return request(app)
    .get(`/group/${groupId}/edit-group-delivery-location`)
    .expect(200)
    .expect(res => {
      expect(res.text).toContain(`/group/${groupId}/group-details`)
    })
})

server/createGroup/location/createOrEditGroupLocationPresenter.ts
```diff
@@ -3,22 +3,34 @@ import { FormValidationError } from '../../utils/formValidationError'
...
     readonly locations: CodeDescription[],
     private readonly validationError: FormValidationError | null = null,
     readonly createGroupFormData: Partial<CreateGroupRequest> | null = null,
     private readonly userInputData: Record<string, unknown> | null = null,
+    private readonly groupId: string | null = null,
+    readonly isEdit = false,
+    private readonly backLink: string | null = null,
   ) {}
 
   get backLinkUri() {
```
raby, line 17: We could add a safe fallback for backLinkUri here when backLink is missing? In the edit flow this value comes from session state, so a direct hit to the route or a fresh session could leave us rendering null/undefined as the Back link target. Falling back to group details, or to the edit PDU step, would make this page much more resilient.

get backLinkUri() {
  if (!this.isEdit) {
    return '/group/create-a-group/group-probation-delivery-unit'
  }

  return this.backLink || `/group/${this.groupId}/group-details`
}

Or, if you want to keep the logic simpler:

get backLinkUri() {
  return this.isEdit
    ? this.backLink || `/group/${this.groupId}/group-details`
    : '/group/create-a-group/group-probation-delivery-unit'
}

rgforsyth (approved):

PR #455, opened 2026-04-16
add success message banner when a cohort or gender of a group is updated

rgforsyth:
add success message banner when a cohort or gender of a group is updated


niaevans9 (approved):
lgtm

PR #457, opened 2026-04-16
Apg 1992/change staff

Alison-Weatherston:
https://accredited-programmes-manage-and-deliver-dev.hmpps.service.justice.gov.uk/group/466ba4c3-2dbd-4cac-bba1-63127af52a5c/group-details

Change links to be added to 'Staff' 
 
Screenshot 2026-04-15 at 16 27 19

Figma Link
https://www.figma.com/design/Y1h09XHyEI9O8KZoOpvlTS/Accredited-Programmes--Community-?node-id=10551-52136&t=9Q2mj7XTQG7BRC4p-1


server/createGroup/editGroupController.ts
```diff
@@ -425,4 +427,71 @@ export default class EditGroupController extends BaseController {
...
+        res.status(400)
+        formError = data.error
+        userInputData = req.body
+      } else {
+        await this.accreditedProgrammesManageAndDeliverService.updateGroup(username, groupId, {
+          teamMembers: data.paramsForUpdate.teamMembers,
+        })
+
+        req.session.createGroupFormData = {}
+        return res.redirect(`/group/${groupId}/group-details`)
```
raby, line 463: Really clean addition overall. One small UX consistency question: should this redirect include a success message query param like the other edit endpoints in this controller? It might help users get the same confirmation pattern after saving facilitator changes.

server/createGroup/editGroupController.ts
```diff
@@ -425,4 +427,71 @@ export default class EditGroupController extends BaseController {
...
+    let formError: FormValidationError | null = null
+
+    const groupDetails = await this.accreditedProgrammesManageAndDeliverService.getGroupDetailsById(username, groupId)
+    const pduMembers = await this.accreditedProgrammesManageAndDeliverService.getPduMembers(username)
+
+    const mapMember = (
+      personName: string,
+      teamMemberType: CreateGroupTeamMember['teamMemberType'],
+    ): CreateGroupTeamMember | null => {
+      const matchingMember = pduMembers.find(member => member.personName === personName)
```
raby: Nice work on this new edit flow. Could I check intended behaviour here: we map existing assigned staff by person name and remove non-matches before submitting. For example, if John Smith is already assigned to the group but is no longer returned in the current PDU members list (team move, rename, or temporary data sync issue), then opening this page and saving could remove John unintentionally. Is that expected, or should previously assigned staff be preserved unless the user explicitly removes them?

server/createGroup/editGroupController.ts
Alison-Weatherston: Yes, it should, but in this instance, the success message is in a separate ticket.

server/views/createGroup/createGroupTreatmentManager.njk
```diff
@@ -282,22 +282,31 @@
...
-
             <div class="moj-button-action">
               {{ govukButton({
               id: "add-another-cover-facilitator",
               text: "Add another cover facilitator",
               classes: "govuk-button--secondary moj-add-another__add-button govuk-!-margin-bottom-4"
               }) }}
+                {% if  isEditJourney %}
+         {{ govukInsetText({
+              text: "This will change the overall group details and facilitators assigned to future sessions. The assigned facilitators will not change for any sessions that have already taken place.",
```
niaevans9: Ideally this should be pulled in from the view

server/createGroup/editGroupController.ts
```diff
@@ -425,4 +427,71 @@ export default class EditGroupController extends BaseController {
     const view = new CreateOrEditGroupLocationView(presenter)
     return this.renderPage(res, view)
   }
+
+  async editGroupTreatmentManager(req: Request, res: Response): Promise<void> {
```
niaevans9, line 432: We need some tests for this controller

server/createGroup/createGroupController.ts
```diff
@@ -289,13 +289,15 @@ export default class CreateGroupController extends BaseController {
     }
 
     const pduMembers = await this.accreditedProgrammesManageAndDeliverService.getPduMembers(username)
-    const presenter = new CreateGroupTreatmentManagerPresenter(
+    const presenter = new CreateOrEditGroupTreatmentManagerPresenter(
+      '',
```
niaevans9: It would be nicer to handle this in the presenter than pass empty strings from the controller

server/routes/index.ts
```diff
@@ -292,6 +292,9 @@ export default function routes({ accreditedProgrammesManageAndDeliverService }:
   getOrPost('/group/create-a-group/group-facilitators', async (req, res) => {
     await createGroupController.showCreateGroupTreatmentManager(req, res)
   })
+  getOrPost('/group/:groupId/edit-group-facilitators', async (req, res) => {
+    await editGroupController.editGroupTreatmentManager(req, res)
+  })
```
thomaswilsonxyz, line 297: This route/path doesn't quite feel right (if it's specified in a sigma, can we ask to push back on it if possible)

I'd expect to see this as:

`/group/:group_id/group-facilitators` and the verb (`GET` vs `POST`, perhaps vs. `PUT`) would let us know if we're viewing or editing

server/createGroup/editGroupController.ts
```diff
@@ -425,4 +427,71 @@ export default class EditGroupController extends BaseController {
...
+    const { groupId } = req.params
+    const { username } = req.user
+    let userInputData = null
+    let formError: FormValidationError | null = null
+
+    const groupDetails = await this.accreditedProgrammesManageAndDeliverService.getGroupDetailsById(username, groupId)
```
thomaswilsonxyz, line 438: It really feels like this should be the responsibility of a BFF endpoint (pulling data from multiple sources, formatting it in a specific way) - is there a reason we're doing it in a controller here ?

server/createGroup/editGroupController.ts
```diff
@@ -425,4 +427,71 @@ export default class EditGroupController extends BaseController {
...
+      ...((groupDetails?.coverFacilitators || []).map(name => mapMember(name, 'COVER_FACILITATOR')) || []),
+    ].filter((member): member is CreateGroupTeamMember => member !== null)
+
+    const createGroupFormData = {
+      groupCode: groupDetails?.code,
+      teamMembers: existingTeamMembers,
+    }
+
+    if (req.method === 'POST') {
+      const data = await new CreateOrEditGroupForm(req, '').createOrEditGroupTreatmentManagerData()
```
thomaswilsonxyz: What's this second argument (`""`) doing here ? other places we call `CreateOrEditGroupForm` looks like we pass just one arg

server/createGroup/createGroupController.test.ts
```diff
@@ -133,6 +133,19 @@ describe('Create Group Controller', () => {
         })
     })
 
+    it('redirects to start date page when group code lookup returns 404', async () => {
```
thomaswilsonxyz: I'm not 100% following this test, I think it might be because it's very edge case-y, but I just want to make sure.

If I attempt to find a group with a code, we pretend the server returns `404` ("no group with that code found")

Then we hit the `POST` endpoint (presumably to create a group, which will then create the group, and attempt to find the group it just created, to redirect the user there)

Are we modelling the edge case where the user 

1. Successfully creates a group using the form 
2. The server `404`s when we ask for the group we just created (and presumably got back the `id` of in the API call we made a request for)
3. The user is redirected back to the "create a group" form 

if this is the case, I think it might make a little more sense to send them back to the Group List page, because (to our knowledge) we have successfully created the group, and sending them back to the Create Group form could cause them to re-create something they just created?

server/createGroup/editGroupController.ts
Alison-Weatherston: test added

server/routes/index.ts
Alison-Weatherston: Spinning up another ticket for this.

niaevans9 (approved):
lgtm

PR #460, opened 2026-04-17
Apg 1729/delete button remove

Alison-Weatherston:
The delete button to be hidden when the start date and time have passed.

Example: https://accredited-programmes-manage-and-deliver-dev.hmpps.service.justice.gov.uk/group/466ba4c3-2dbd-4cac-bba1-63127af52a5c/session/cb947ab1-8a9d-47d2-bc9d-92b123b3c71b/edit-session

Screenshot 2026-04-17 at 12 29 40


server/editSession/editSessionPresenter.ts
```diff
@@ -26,8 +26,80 @@ export default class EditSessionPresenter {
     }
   }
 
+  private monthNameToNumber(monthName: string): number | null {
```
niaevans9: I think we already do something like this in the groupDetailsPresenter - might be worth having a look if we can re-use it and pull it out into a utility class?

Alison-Weatherston: We do. I have moved the duplicate code into the DateUtils file and called that from both.

niaevans9 (approved):
lgtm

raby (commented):
LGTM

PR #469, opened 2026-04-21
APG-1968 Add LDC score to Learning Needs page

pfcairns:
This pull request updates the Learning Needs page to include the LDC score in line with the equivalent API updates


niaevans9 (approved):
lgtm

PR #471, opened 2026-04-22
heading name updated from Active alerts to NDelius risk flags (regist…

Alison-Weatherston:
Heading 'Active alerts' updated to 'NDelius risk flags (registers)' and the text changed from 'Nomis' to 'NDelius in the standout text'.

Screenshot 2026-04-22 at 09 07 49


niaevans9 (approved):
lgtm

PR #475, opened 2026-04-22
APG-1968 Replace `scoreValueText` with `PresenterUtils.scoreValueText`

pfcairns:
- Fix to handle a null ldcScore value from the API appropriately
- Refactored all instances of `scoreValueText` in presenters and tests to use `PresenterUtils.scoreValueText`, ensuring consistency and code reusability across the codebase.

### DIFF
--- a/server/pni/pniPresenter.test.ts
+++ b/server/pni/pniPresenter.test.ts
@@ -1,5 +1,6 @@
 import { PniScore, ReferralDetails } from '@manage-and-deliver-api'
 import PniPresenter from './pniPresenter'
+import PresenterUtils from '../utils/presenterUtils'
 import pniScoreFactory from '../testutils/factories/pniScoreFactory'
 import referralDetailsFactory from '../testutils/factories/referralDetailsFactory'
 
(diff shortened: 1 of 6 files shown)


raby:
Great fix and nice refactor pulling scoreValueText into a shared static helper. One small test gap: the existing learning needs test only asserts that `ldcScore` renders when it has a value `ldcScore.toString()`, but there is no test covering the null case that this PR is specifically fixing. 

Could we add a test alongside the existing "minimal data" test that passes `ldcScore`: null and asserts the page renders Score missing? That would lock in the fix and match the same null/undefined coverage already in [presenterUtils.test.ts]at [presenterUtils.test.ts:375].

Example in the same style as the existing minimal data test at [risksAndNeedsController.test.ts:107]:

```
it('renders Score missing when ldcScore is null', async () => {
  const learningNeeds: LearningNeeds = learningNeedsFactory.build({ ldcScore: null })
  accreditedProgrammesManageAndDeliverService.getLearningNeeds.mockResolvedValue(learningNeeds)

  const referralId = randomUUID()
  return request(app)
    .get(`/referral/${referralId}/learning-needs`)
    .expect(200)
    .expect(res => {
      expect(res.text).toContain('Score missing')
    })
})
```

pfcairns:
> Great fix and nice refactor pulling scoreValueText into a shared static helper. One small test gap: the existing learning needs test only asserts that `ldcScore` renders when it has a value `ldcScore.toString()`, but there is no test covering the null case that this PR is specifically fixing.
> 
> Could we add a test alongside the existing "minimal data" test that passes `ldcScore`: null and asserts the page renders Score missing? That would lock in the fix and match the same null/undefined coverage already in [presenterUtils.test.ts]at [presenterUtils.test.ts:375].
> 
> Example in the same style as the existing minimal data test at [risksAndNeedsController.test.ts:107]:
> 
> ```
> it('renders Score missing when ldcScore is null', async () => {
>   const learningNeeds: LearningNeeds = learningNeedsFactory.build({ ldcScore: null })
>   accreditedProgrammesManageAndDeliverService.getLearningNeeds.mockResolvedValue(learningNeeds)
> 
>   const referralId = randomUUID()
>   return request(app)
>     .get(`/referral/${referralId}/learning-needs`)
>     .expect(200)
>     .expect(res => {
>       expect(res.text).toContain('Score missing')
>     })
> })
> ```

Good spot Raby, I had meant to include this addtional test. Will address.

raby (approved):
LGTM - Nice additional test

PR #485, opened 2026-04-24
Maintain the previous input value when there is a form error

niaevans9:
Maintain the previous input value when there is a form error on the edit group date page


raby (approved):
LGTM

PR #489, opened 2026-04-28
Attendance History Tweaks

niaevans9:
- Correct date ordering
- Remove session name ordering - sessions are being returned from the API in date order
- Capitalise first letter of time value (e.g. Midday to 1pm)
- Implement changes due to types update of facilitators object


Alison-Weatherston (approved):
LGTM

PR #494, opened 2026-04-30
Delete session button logic updates

niaevans9:
The delete button should NOT appear if:
It is a core group session
OR The session is in the past
OR It has any attendance recorded against it

### DIFF
--- a/server/editSession/editSessionPresenter.test.ts
+++ b/server/editSession/editSessionPresenter.test.ts
@@ -587,7 +587,7 @@ describe('EditSessionPresenter', () => {
       expect(presenter.canBeDeleted).toBe(false)
     })
 
-    it('returns true when no referrals are allocated, the session is in the past and is a group catch up', () => {
+    it('returns true when no referrals are allocated, the session is in the future and is a group catch up', () => {
       const presenter = new EditSessionPresenter(
         mockGroupId,
         buildSessionDetails({
--- a/server/editSession/editSessionPresenter.test.ts
+++ b/server/editSession/editSessionPresenter.test.ts
@@ -445,7 +445,7 @@ describe('EditSessionPresenter', () => {
       sessionType: 'Individual',
       isCatchup: false,
       attendanceAndSessionNotes: [],
-      date: 'Sunday 17 May 2026',
+      date: 'Sunday 17 May 2126',
       time: '11:57am to 11:58am',
       scheduledToAttend: [],
       facilitators: [],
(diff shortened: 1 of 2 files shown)


server/editSession/editSessionPresenter.ts
```diff
@@ -72,16 +72,20 @@ export default class EditSessionPresenter {
   }
 
   get canBeDeleted(): boolean {
```
raby, line 74: Really nice improvement to the delete-button rule set here. I like how the logic now reflects the actual business intent (core group sessions blocked, past sessions blocked, attendance-recorded sessions blocked) instead of relying on session type alone. This makes the behaviour easier to understand.

server/editSession/editSessionPresenter.ts
```diff
@@ -72,16 +72,20 @@ export default class EditSessionPresenter {
...
-    if (this.sessionDetails.sessionType.toLowerCase() !== 'individual') {
+    // Cant be deleted if its a core group session
+    if (this.sessionDetails.sessionType.toUpperCase() === 'GROUP' && this.sessionDetails.isCatchup === false) {
       return false
     }
-
-    const sessionStart = this.sessionStartDateTime()
-    if (sessionStart === null) {
+    // Cant be deleted if its in the past
+    if (this.sessionStartDateTime() === null || this.sessionStartDateTime().getTime() < Date.now()) {
```
raby: Probably a boundary condition never achieved in reality due to milliseconds -however using < Date.now() means a session is still deletable at the exact start instant? On main, this behaved as start must be strictly in the future just had a check. Should this be <= Date.now() so deletion is blocked once the session starts? (just in case)

server/editSession/editSessionPresenter.test.ts
```diff
@@ -445,39 +445,161 @@ describe('EditSessionPresenter', () => {
...
+          ],
+        }),
         mockSessionId,
         mockDeleteUrl,
       )
 
       expect(presenter.canBeDeleted).toBe(false)
     })
+
+    it('returns true when no referrals are allocated, the session is in the past and is a group catch up', () => {
```
raby: Possible test mismatch: this test says “session is in the past,” but it appears to rely on the shared default date (2126), which makes it future unless overridden. Could we set an explicit past date/time in this test so it validates the scenario described in the title? Cheers :-) 

buildSessionDetails({
  sessionType: 'Group',
  isCatchup: true,
  attendanceAndSessionNotes: [],
  date: 'Sunday 17 May 2026',
  time: '11:57am to 11:58am',
})

Alison-Weatherston:
Tested locally, and it works well.

server/editSession/editSessionPresenter.test.ts
niaevans9: Good spot, typo

raby (approved):
Great!

PR #498, opened 2026-04-30
allow creation of a group in the past if in specific regions

rgforsyth:
allow creation of a group in the past if in specific regions.
Updated helm files to set new environment variable.

N50MANC has been added to the allowlist temporarily for testing. This is Cleveland, our local users region.

### DIFF
--- a/server/config.ts
+++ b/server/config.ts
@@ -56,7 +56,8 @@ export default {
   production,
   https: process.env.NO_HTTPS === 'true' ? false : production,
   staticResourceCacheDuration: '1h',
-  enable_past_date_region_restriction: get('ENABLE_PAST_DATE_REGION_RESTRICTION', 'false') === 'true',
+  enable_past_date_region_restriction:
+    get('ENABLE_PAST_DATE_REGION_RESTRICTION', 'false', requiredInProduction) === 'true',
   redis: {
     enabled: get('REDIS_ENABLED', 'false', requiredInProduction) === 'true',
     host: get('REDIS_HOST', 'localhost', requiredInProduction),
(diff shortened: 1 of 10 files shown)


server/createGroup/allowedPastDateRegions.ts
```diff
@@ -0,0 +1,21 @@
+export const ALLOWED_PAST_DATE_REGIONS: string[] = ['N50MANC']
```
rgforsyth, line 1: N50MANC has been added to the allowlist temporarily for testing. This is Cleveland, our local users region.

server/config.ts
```diff
@@ -56,6 +56,7 @@ export default {
   production,
   https: process.env.NO_HTTPS === 'true' ? false : production,
   staticResourceCacheDuration: '1h',
+  enable_past_date_region_restriction: get('ENABLE_PAST_DATE_REGION_RESTRICTION', 'false') === 'true',
```
Jack-Dempster:
```suggestion
  enable_past_date_region_restriction: get('ENABLE_PAST_DATE_REGION_RESTRICTION', 'false', requiredInProduction) === 'true',
```
might want to add the requiredInProduction flag

server/createGroup/allowedPastDateRegions.ts
Jack-Dempster, line 1: are we planning on managing the regions by env variables in the future?

rgforsyth: I believe Wilson initially said we will be changing it pretty infrequently so just store within the application for now.

server/config.ts
rgforsyth: Good spot

Jack-Dempster (approved):
👍

PR #504, opened 2026-05-05
APG-2136 Update rescheduling heading and hint text for session rescheduling 

pfcairns:
### Description

Updated terminology and guidance related to rescheduling in edit session views in line with Figma updates


raby (approved):
LGTM

PR #506, opened 2026-05-05
Remove hardcoding of home link in header

Jack-Dempster:
The banner hmpps link was always pointing to the Prod url. This PR reads the URL from the env variables of the env it's in and sets the url accordingly.


niaevans9 (approved):
lgtm

PR #508, opened 2026-05-06
pull region information, rather than pdu

rgforsyth:
Use the users region rather than pdu for session storage.


niaevans9 (approved):
lgtm

PR #511, opened 2026-05-07
Cant delete sessions once the end date has passed

niaevans9:
Change of logic for deleting a session
You can now deleted a session when the start date/time has passed
But you cant delete it once the end date/time has passed


rgforsyth (approved):

raby (commented):
looks good

PR #512, opened 2026-05-07
APG-2088-text updates

Alison-Weatherston:
Comments from the ticket:

When changing the cohort via a PoP space, there is a screen ‘Which cohort should [Latoya Barton] be in?’
There is a content change on the page:
Remove the comma on the inset text playback: “Latoya Barton's, current cohort:” to “Latoya Barton's current cohort:”
There also isn’t an h1 on the page:
Change “Which cohort should [Latoya Barton] be in?” to be an h1, not an h2

**Changes**
1. comma removed from 'Colin Maggio-Schimmel's, current cohort:'
2. Which cohort should Colin Maggio-Schimmel be in? - changed to be an h1

Before

Screenshot 2026-05-07 at 10 06 22

After

Screenshot 2026-05-07 at 10 09 11

### DIFF
--- a/server/cohort/changeCohortView.ts
+++ b/server/cohort/changeCohortView.ts
@@ -20,7 +20,7 @@ export default class ChangeCohortView {
 
   get currentCohortText(): InsetTextArgs {
     return {
-      html: `<p> ${this.presenter.details.personName}'s, current cohort: </p>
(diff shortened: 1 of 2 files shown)


niaevans9 (approved):
lgtm

PR #532, opened 2026-05-19
feat: Add group size reporting endpoint

thomaswilsonxyz:
**WHAT**

This PR introduces a new façade wrapper around the `GET /reporting/group-size.csv` endpoint on the API.  

**WHY**

As part the the reporting architecture (outlined in a [doc here](https://github.com/ministryofjustice/hmpps-accredited-programmes-manage-and-deliver-api/blob/main/docs/2026-05-reporting.md)) we are exposing Group Size reporting through an endpoint which allows the user to download a .csv file. 

This is point-in-time, not event-driven, reporting, and therefore event oriented system (e.g. App Insights) would not be appropriate.

**HOW**

* Generated the API types from the new endpoint (`index.d.ts`)
* Created the `ReportingController` which is a simple façade around the API endpoints

### DIFF
--- a/server/reporting/reportingController.ts
+++ b/server/reporting/reportingController.ts
@@ -31,9 +31,15 @@ export default class ReportingController {
     }
 
     const { username } = req.user
-    const csv = await this.accreditedProgrammesManageAndDeliverService.getGroupSizeReport(username, groupStartedSince)
+    const { csv, headers } = await this.accreditedProgrammesManageAndDeliverService.getGroupSizeReport(
+      username,
+      groupStartedSince,
+    )
 
     res.setHeader('Content-Type', 'text/csv; charset=utf-8')
+    Object.entries(headers).forEach(([key, value]) => {
+      res.setHeader(key, value)
+    })
     res.send(csv)
   }
 }
--- a/server/reporting/reportingController.ts
+++ b/server/reporting/reportingController.ts
@@ -1,0 +1,39 @@
+import { Request, Response } from 'express'
+
+import AccreditedProgrammesManageAndDeliverService from '../services/accreditedProgrammesManageAndDeliverService'
+
+export default class ReportingController {
+  constructor(
+    private readonly accreditedProgrammesManageAndDeliverService: AccreditedProgrammesManageAndDeliverService,
+  ) {}
+
+  private static hasValidDateString(value: unknown): value is string {
+    if (typeof value !== 'string') {
+      return false
+    }
+
+    const parsedDate = new Date(value)
+    if (Number.isNaN(parsedDate.getTime())) {
+      return false
+    }
+
+    // Ensure the parsed value can be serialised as an ISO date string.
+    parsedDate.toISOString()
+    return true
+  }
+
+  async downloadGroupSizeReport(req: Request, res: Response): Promise<void> {
+    const { groupStartedSince } = req.query
+
+    if (!ReportingController.hasValidDateString(groupStartedSince)) {
+      res.status(400).send('groupStartedSince must be a valid date-time string')
+      return
+    }
+
+    const { username } = req.user
+    const csv = await this.accreditedProgrammesManageAndDeliverService.getGroupSizeReport(username, groupStartedSince)
+
+    res.setHeader('Content-Type', 'text/csv; charset=utf-8')
+    res.send(csv)
+  }
+}
--- a/server/@types/manageAndDeliverApi/imported/index.d.ts
+++ b/server/@types/manageAndDeliverApi/imported/index.d.ts
@@ -352,6 +352,22 @@ export interface paths {
     patch?: never
     trace?: never
   }
+  '/dev/seed/referrals': {
+    parameters: {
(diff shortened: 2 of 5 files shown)


server/reporting/reportingController.ts
```diff
@@ -0,0 +1,39 @@
...
+
+    if (!ReportingController.hasValidDateString(groupStartedSince)) {
+      res.status(400).send('groupStartedSince must be a valid date-time string')
+      return
+    }
+
+    const { username } = req.user
+    const csv = await this.accreditedProgrammesManageAndDeliverService.getGroupSizeReport(username, groupStartedSince)
+
+    res.setHeader('Content-Type', 'text/csv; charset=utf-8')
```
raby, line 39: Currently, the CSV response only sets Content-Type but browsers need Content-Disposition to properly handle file downloads.

LLM:
` const timestamp = new Date().toISOString().split('T')[0]`

Me:
`  res.setHeader('Content-Disposition', `attachment; filename="group-size-report-${timestamp}.csv"`)`

This ensures the Browser prompts download instead of displaying content and Suggested filename includes date for organsiation but could be another method depending on user need

server/reporting/reportingController.ts
```diff
@@ -0,0 +1,39 @@
...
+  async downloadGroupSizeReport(req: Request, res: Response): Promise<void> {
+    const { groupStartedSince } = req.query
+
+    if (!ReportingController.hasValidDateString(groupStartedSince)) {
+      res.status(400).send('groupStartedSince must be a valid date-time string')
+      return
+    }
+
+    const { username } = req.user
+    const csv = await this.accreditedProgrammesManageAndDeliverService.getGroupSizeReport(username, groupStartedSince)
```
raby: The service call should be wrapped in try-catch to handle potential API failures gracefully.

LLM:

` try {
    const { username } = req.user
    const csv = await this.accreditedProgrammesManageAndDeliverService.getGroupSizeReport(username, groupStartedSince)

    const timestamp = new Date().toISOString().split('T')[0]
    res.setHeader('Content-Type', 'text/csv; charset=utf-8')
    res.setHeader('Content-Disposition', `attachment; filename="group-size-report-${timestamp}.csv"`)
    res.send(csv)
  } catch (error) {
    logger.error('Error generating group size report', error)
    res.status(500).send('Unable to generate report. Please try again later.')
  }`

raby: Test for error by LLM:

`it('returns 500 when service throws an error', async () => {
  accreditedProgrammesManageAndDeliverService.getGroupSizeReport.mockRejectedValue(
    new Error('Service unavailable')
  )
  
  app = appWithAllRoutes({
    services: { accreditedProgrammesManageAndDeliverService },
    userSupplier: () => ({
      ...defaultUser,
      token: createUserToken(['ROLE_ACCREDITED_PROGRAMMES_MANAGE_AND_DELIVER_API__ACPMAD_UI_REPORTING']),
    }),
  })

  await request(app)
    .get('/reporting/group-size.csv?groupStartedSince=2026-05-18T13:30:00')
    .expect(500)
    .expect(res => {
      expect(res.text).toBe('Unable to generate report. Please try again later.')
    })
})`

server/reporting/reportingController.ts
```diff
@@ -0,0 +1,39 @@
...
+    // Ensure the parsed value can be serialised as an ISO date string.
+    parsedDate.toISOString()
+    return true
+  }
+
+  async downloadGroupSizeReport(req: Request, res: Response): Promise<void> {
+    const { groupStartedSince } = req.query
+
+    if (!ReportingController.hasValidDateString(groupStartedSince)) {
+      res.status(400).send('groupStartedSince must be a valid date-time string')
```
raby, line 29: Small improvement for developer troubleshooting, the current validation handles both missing and invalid dates with the same error message. Consider adding a more specific check?

`if (!groupStartedSince) {
  res.status(400).send('groupStartedSince query parameter is required')
  return
}

if (!ReportingController.hasValidDateString(groupStartedSince)) {
  res.status(400).send('groupStartedSince must be a valid date-time string')
  return
}`

thomaswilsonxyz: Probably overkill for now, we can add in later if we need :)

server/reporting/reportingController.ts
thomaswilsonxyz: Good shout, i've passed along the headers that the API generates - to centralise them there

server/reporting/reportingController.ts
thomaswilsonxyz: I've implemented this, but in another way

server/reporting/reportingController.ts
raby: Sure yes I had similar thoughts re overkill

server/reporting/reportingController.ts
raby: Great

rgforsyth (approved):

PR #533, opened 2026-05-19
page title and URLs updated for groups list

Alison-Weatherston:
URL and page titles updated for the groups list page.

Screenshot 2026-05-19 at 10 36 43

### DIFF
--- a/server/group/groupPresenter.ts
+++ b/server/group/groupPresenter.ts
@@ -61,6 +61,12 @@ export default class GroupPresenter {
     return this.section === GroupListPageSection.COMPLETE ? 'Completed groups' : 'Not started or in progress groups'
   }
 
+  get pageTitle(): string {
+    return this.section === GroupListPageSection.COMPLETE
+      ? 'Completed Building Choices groups'
+      : 'Not started and in progress Building Choices groups'
+  }
+
   get groupTableArgs(): TableArgs {
     return {
       caption: this.tableCaption,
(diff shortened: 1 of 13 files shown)


server/group/groupPresenter.ts
```diff
@@ -175,7 +181,7 @@ export default class GroupPresenter {
       items: [
         {
           text: `Not started or in progress (${this.section === GroupListPageSection.NOT_STARTED_OR_IN_PROGRESS ? this.groupListItems.totalElements : this.otherGroupListCountTotal})`,
```
raby, line 183: Hi, don't know if a design choice ignore if so however I noticed the sub-navigation tab text says "Not started or in progress" but the URL and page title both use "and". This creates an inconsistency?

Before:
`{
  text: `Not started or in progress (${...})`,  // "or"
  href: possiblyAppendParams(`/groups/not-started-and-in-progress`, this.params),  // "and"
  active: this.section === GroupListPageSection.NOT_STARTED_OR_IN_PROGRESS,
},`

After a suggestion if valid bug:

`{
  text: `Not started and in progress (${...})`,  // "and"
  href: possiblyAppendParams(`/groups/not-started-and-in-progress`, this.params),
  active: this.section === GroupListPageSection.NOT_STARTED_OR_IN_PROGRESS,
},`

groupPresenter.ts (line 183)
groupPresenter.test.ts (lines 25, 56, 86)

Why "and" is better:

Grammatically correct (a group can be both not started AND currently in progress)
Matches the page title and URL
More precise and consistent

Alison-Weatherston: @raby Yes, this is correct. I asked Flora the same question, and that is what they want.

raby: Thank you for clarifying

raby (approved):

PR #542, opened 2026-05-21
Apg 2258/update page titles urls edit sessions

Alison-Weatherston:
URLs and page titles for 'Edit session'

Example page:

Screenshot 2026-05-21 at 09 34 15


raby (commented):
PageTitle is used consistently across the codebase, one trace through is using pageTitles possibly accident. 

Had a look, a good llm explanation: 

"bug is invisible to TypeScript and tests because the wrong name is used **consistently** across the presenter getter, the view, and the Nunjucks template — so nothing errors. It only surfaces at runtime as `"undefined - Application Name"` in the browser tab.

### Why it's easy to miss
Nunjucks templates don't throw if a variable is undefined — they silently render `"undefined"`. TypeScript also won't catch this because the view passes a plain object literal (`{ pageTitles: ... }`), and the template just reads whatever variable name is in scope."

server/editSession/facilitators/editSessionFacilitatorsPresenter.test.ts
```diff
@@ -0,0 +1,286 @@
...
+
+      expect(backLinkArgs).toEqual({
+        text: 'Back',
+        href: linkUrl,
+      })
+    })
+  })
+
+  describe('pageTitles', () => {
+    it('should return Edit the session facilitators', () => {
```
raby, line 48: 'should return page title'?

server/editSession/facilitators/editSessionFacilitatorsPresenter.test.ts
```diff
@@ -0,0 +1,286 @@
...
+      const presenter = new EditSessionFacilitatorsPresenter(linkUrl, groupId, editSessionFacilitatorsResponse)
+
+      expect(presenter.pageTitles).toBe('Edit the session facilitators')
```
raby, line 51: expect(presenter.pageTitle).toBe('Edit the session facilitators')

server/editSession/facilitators/editSessionFacilitatorsPresenter.ts
```diff
@@ -19,6 +19,10 @@ export default class EditSessionFacilitatorsPresenter {
     }
   }
 
+  get pageTitles(): string {
```
raby, line 22: pageTitle()

server/editSession/facilitators/editSessionFacilitatorsView.ts
```diff
@@ -64,6 +57,7 @@ export default class EditSessionFacilitatorsView {
       'editSession/editSessionFacilitators',
       {
         backLinkArgs: this.presenter.backLinkArgs,
+        pageTitles: this.presenter.pageTitles,
```
raby, line 60: pageTitle:

server/views/editSession/editSessionFacilitators.njk
```diff
@@ -5,7 +5,7 @@
 {% from "govuk/components/select/macro.njk" import govukSelect %}
 {%- from "govuk/components/fieldset/macro.njk" import govukFieldset -%}
 
-{% set pageTitle = applicationName + " - Home" %}
+{% set pageTitle = pageTitles + " - " + applicationName %}
```
raby, line 8: {% set pageTitle = pageTitle + " - " + applicationName %}

server/editSession/facilitators/editSessionFacilitatorsPresenter.test.ts
Alison-Weatherston: Thanks @Raby - in this instance pageTitles is correct because the API returns 'pageTitle' for another variable on this page so i had to and the 's' here.

server/editSession/facilitators/editSessionFacilitatorsPresenter.ts
Alison-Weatherston: meant to be this for facilitators

server/editSession/facilitators/editSessionFacilitatorsView.ts
Alison-Weatherston: as above

raby (approved):

PR #547, opened 2026-05-22
prevent lengthening of past session

Alison-Weatherston:


Alison-Weatherston:
Closing to recreate as requested.

PR #548, opened 2026-05-22
missing page title on /add-motivation-background-and-non-associations page

Alison-Weatherston:
I missed a page title on the following page: https://accredited-programmes-manage-and-deliver-dev.hmpps.service.justice.gov.uk/referral/f1fd3bc3-bdfb-4807-b6ad-eb66fd28eb57/add-motivation-background-and-non-associations


raby (commented):
Great work on this one! The `pageTitle` is correctly wired through all three layers (presenter → view → template), and the new test file goes well beyond the minimum — covering field population, edge cases, validation errors and error summary formatting. Really solid test coverage, especially for a catch-up PR.

One minor thing to confirm: the page title currently renders as a three-part browser tab title:

`Motivation, background and non-associations - Availability and motivation - Application Name`

Most other pages in this flow use a two-part format (`Page name - Application Name`). 

If design has specified the section name should be included here then this is absolutely fine — just worth a quick check so we're consistent across the flow.

raby (approved):

PR #557, opened 2026-05-28
feat: Add several more reporting endpoints to the report

thomaswilsonxyz:
**WHAT**

This commit does several things:

1. Add additional reporting endpoints
2. Remove reference to the ROLE_ACCREDITED_PROGRAMMES_MANAGE_AND_DELIVER_API__ACPMAD_UI_REPORTING
3. Bumps the backend types (from the latest version of `main`)

**WHY**

Creating facades from the UI through to the remaining two Reporting endpoints allows us to collect the reporting data we
expect to be useful.

At time of writing we have not successfully had the HMPPS auth team create the role, and therefore we are unable to test
both reporting and onboarding.  It is currently critical that we are able to test both of these functions in production
in the coming week.


server/reporting/reportingController.ts
```diff
@@ -22,24 +23,173 @@ export default class ReportingController {
...
+  }
+
+  private static readOptionalStringQueryParam(value: unknown): string | undefined {
+    return typeof value === 'string' ? value : undefined
+  }
+
+  private static logReportRequest(req: Request, reportName: string, query: Record<string, string | undefined>): void {
+    logger.info(
+      {
+        username: req.user.username,
```
raby, line 42: An edge case: req.user can be undefined if the auth middleware hasn't populated it (e.g. misconfigured middleware order, test environments). Accessing .username directly will throw TypeError: Cannot read properties of undefined, crashing the request.

`username: req.user?.username,`

server/reporting/reportingController.ts
```diff
@@ -22,24 +23,173 @@ export default class ReportingController {
...
+    if (typeof value !== 'string' || !/^\d{4}-\d{2}-\d{2}$/.test(value)) {
+      return false
+    }
+
+    const parsedDate = new Date(`${value}T00:00:00.000Z`)
+    return !Number.isNaN(parsedDate.getTime()) && parsedDate.toISOString().startsWith(value)
+  }
+
+  private static readOptionalStringQueryParam(value: unknown): string | undefined {
+    return typeof value === 'string' ? value : undefined
```
raby, line 36: Express parses duplicate query params as an array (e.g. ?referralsCreatedSince=a&referralsCreatedSince=b → ['a', 'b']). 

The current implementation returns undefined for arrays, which means the "at least one filter required" check still fires correctly — but the caller gets no feedback that their input was malformed. If a user accidentally sends an array for a required param, it's silently dropped.

Proposed:
`private static readOptionalStringQueryParam(value: unknown): string | undefined {
  if (Array.isArray(value)) {
    // Express parses duplicate query params as arrays — take first value
    return typeof value[0] === 'string' ? value[0] : undefined
  }
  return typeof value === 'string' ? value : undefined
}`

server/reporting/reportingController.ts
```diff
@@ -22,24 +23,173 @@ export default class ReportingController {
...
+    if (!query.referralsCreatedSince && !query.referralsCompletedAfter) {
+      res.status(400).send('At least one of referralsCreatedSince or referralsCompletedAfter must be provided')
+      return
+    }
+
+    const { username } = req.user
+    ReportingController.logReportRequest(req, 'dosage', query)
+    const { csv, headers } = await this.accreditedProgrammesManageAndDeliverService.getDosageReport(username, query)
+
+    ReportingController.sendCsv(res, csv, headers)
```
raby, line 108: No error handling around service calls — API errors surface as unhandled exceptions

If the downstream API returns a 4xx/5xx, the `RestClient` throws and it bubbles up as an unhandled exception. The global error handler could return an HTML error page, which is wrong for a CSV download endpoint. The caller would get an HTML response instead of an error it can handle.

`try {
  const { csv, headers } = await this.accreditedProgrammesManageAndDeliverService.getDosageReport(username, query)
  ReportingController.sendCsv(res, csv, headers)
} catch (error) {
  logger.error({ error, username, query }, '[ReportingController] Failed to fetch dosage report')
  res.status(502).send('Failed to retrieve report from upstream service')
}`

raby: lines 74, 109, 140, 192 also

script/reporting-url-generator.v1.1.0.html
```diff
@@ -0,0 +1,425 @@
...
+          {
+            title: 'Group size report',
+            path: '/reporting/group-size.csv',
+            fields: [
+              {
+                key: 'groupStartedSince',
+                label: 'Group started since',
+                type: DATETIME_INPUT,
+                defaultValue: date => formatDateTimeLocalInput(date),
+                serialiseValue: value => new Date(value).toISOString(),
```
raby, line 223: `new Date(value).toISOString()` throws `RangeError: Invalid time value `if value is an invalid date string. While it guards against empty strings, a partially typed or browser-cleared datetime input can produce an invalid (non-empty) string that passes the if (!rawValue) check but still causes new Date() to return Invalid Date

// lines 223, 272, 279, 286 (same pattern in each)
`serialiseValue: value => {
  const d = new Date(value)
  return Number.isNaN(d.getTime()) ? '' : d.toISOString()
},`

raby (approved):

PR #558, opened 2026-05-28
No ticket/app insights include username

DanEgan1:


server/utils/azureAppInsights.ts
```diff
@@ -24,18 +25,23 @@ export function buildAppInsightsClient(
   if (process.env.APPLICATIONINSIGHTS_CONNECTION_STRING) {
     defaultClient.context.tags['ai.cloud.role'] = overrideName || applicationName
     defaultClient.context.tags['ai.application.ver'] = buildNumber
+    defaultClient.addTelemetryProcessor((envelope: Contracts.EnvelopeTelemetry, contextObjects) => {
+      const isRequest = envelope?.data?.baseType === Contracts.TelemetryTypeString.Request
+      const username = contextObjects?.['http.ServerRequest']?.res?.locals?.user?.username
```
raby: A small suggestion: We could add a fallback (e.g., 'anonymous') if the username is missing, to ensure the property is always present and avoid undefined/null values in telemetry   

const username = contextObjects?.['http.ServerRequest']?.res?.locals?.user?.username || 'anonymous'

Happy with an improved default other then anonymous.

raby (dismissed):
looks good

server/utils/azureAppInsights.ts
```diff
@@ -24,18 +25,27 @@ export function buildAppInsightsClient(
   if (process.env.APPLICATIONINSIGHTS_CONNECTION_STRING) {
     defaultClient.context.tags['ai.cloud.role'] = overrideName || applicationName
     defaultClient.context.tags['ai.application.ver'] = buildNumber
+    defaultClient.addTelemetryProcessor((envelope: Contracts.EnvelopeTelemetry, contextObjects) => {
+      const isRequest = envelope?.data?.baseType === Contracts.TelemetryTypeString.Request
+
+      if (isRequest && envelope.data && envelope.data.baseData) {
```
thomaswilsonxyz: I think we can optional chain these

```js
if (isRequest && envelope?.data?.baseData) { .. }
```

thomaswilsonxyz (dismissed):

thomaswilsonxyz (approved):

PR #560, opened 2026-06-01
Catch-up text added to attendance and session notes pages. URL and page title

Alison-Weatherston:
Adding 'catch-up' to session notes and attendance page titles and URLs

Screenshot 2026-06-01 at 14 31 37

Screenshot 2026-06-01 at 14 32 02

Screenshot 2026-06-01 at 14 32 22

Screenshot 2026-06-01 at 14 32 41


server/attendance/attendanceController.test.ts
```diff
@@ -128,6 +128,37 @@ describe('showRecordAttendancePages', () => {
...
+      await request(app)
+        .get('/111/6789/record-attendance?referralId=referral2')
+        .expect(302)
+        .expect('Location', '/111/6789/referral/referral2/pre-group-one-to-one-catch-up-session-notes')
+
+      expect(accreditedProgrammesManageAndDeliverService.getRecordAttendanceBffData).toHaveBeenCalledWith(
+        'user1',
+        '6789',
+        ['referral2'],
+      )
```
raby, line 160: The new GET catch-up redirect test is great. I noticed the POST path hits the same notesPageGroupTitle method (line 99 of the controller), but there's no test for it when isCatchup: true. If that method ever regresses the POST flow would silently break.
Could you add something like this alongside the existing POST tests?

`> it('redirects to catch-up session notes URL after POST when single referral ID', async () => {
>   const bffData = recordSessionAttendanceFactory.build({
>     sessionModule: 'Pre-group one-to-one',
>     isCatchup: true,
>   })
>   accreditedProgrammesManageAndDeliverService.getRecordAttendanceBffData.mockResolvedValue(bffData)
>
>   await request(app)
>     .post('/111/6789/record-attendance')
>     .send({ referralIds: ['referral1'], attendees: JSON.stringify([{ referralId: 'referral1', outcomeCode: 'ATTC' }]) })
>     .expect(302)
>     .expect('Location', '/111/6789/referral/referral1/pre-group-one-to-one-catch-up-session-notes')
> })`

raby (dismissed):
Great work, small testy comment

server/attendance/attendanceController.test.ts
Alison-Weatherston: test added

package.json
```diff
@@ -108,6 +108,7 @@
     "jwt-decode": "^4.0.0",
     "moment": "^2.30.1",
     "nocache": "^4.0.0",
+    "npm": "11.16.0",
```
raby: Looks like npm has accidentally been added as a runtime dependency here! npm is a package manager it shouldn't live in dependencies (that's what gets installed in production). This was probably added by mistake when running npm install npm in the terminal?

Could we remove this line and then run npm install to clean up package-lock.json before merging?

Alison-Weatherston: @raby - this has been removed. thank you

raby (approved):
Looks good

PR #571, opened 2026-06-04
Apg 2138/prevent lengthening of past session duration 0306

Alison-Weatherston:
Session where the end date and time are in the past should be able to have the time changed but only if the session is not made any longer.

Help from Claude esp. on the DateUtils.

Error message showing

Screenshot 2026-06-04 at 14 19 53

### DIFF
--- a/server/editSession/dateAndTime/editSessionDateAndTimeForm.ts
+++ b/server/editSession/dateAndTime/editSessionDateAndTimeForm.ts
@@ -63,56 +63,23 @@ export default class EditSessionDateAndTimeFormForm {
     }
   }
 
+  private static startOrEndTimeChanged(
+    originalTime: { hour: number; minutes: number; amOrPm: 'AM' | 'PM' } | undefined,
+    submittedHour: number,
+    submittedMinute: number,
+    submittedPeriod: 'AM' | 'PM',
+  ): boolean {
+    if (!originalTime) return true
+
+    const original = DateUtils.convertTo24Hour(originalTime.hour, originalTime.minutes, originalTime.amOrPm)
+    const submitted = DateUtils.convertTo24Hour(submittedHour, submittedMinute, submittedPeriod)
+
+    return original.hour !== submitted.hour || original.minute !== submitted.minute
+  }
+
   private createSessionDetailsValidations(): ValidationChain[] {
     const { isSessionEnded } = this
-    function durationInMinutes(
-      startHour: number,
-      startMinute: number,
-      startPeriod: 'AM' | 'PM',
-      endHour: number,
-      endMinute: number,
-      endPeriod: 'AM' | 'PM',
-    ): number {
-      const start = DateUtils.convertTo24Hour(startHour, startMinute, startPeriod)
-      const end = DateUtils.convertTo24Hour(endHour, endMinute, endPeriod)
-      return end.hour * 60 + end.minute - (start.hour * 60 + start.minute)
-    }
-
-    function isTimeBefore(
-      startHour: number,
-      startMinute: number,
-      startPeriod: 'AM' | 'PM',
-      endHour: number,
-      endMinute: number,
-      endPeriod: 'AM' | 'PM',
-    ): boolean {
-      const start = DateUtils.convertTo24Hour(startHour, startMinute, startPeriod)
-      const end = DateUtils.convertTo24Hour(endHour, endMinute, endPeriod)
-
-      if (start.hour !== end.hour) {
-        return start.hour < end.hour
-      }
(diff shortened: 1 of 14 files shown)


server/editSession/editSessionController.ts
```diff
@@ -21,16 +21,135 @@ import EditSessionFacilitatorsView from './facilitators/editSessionFacilitatorsV
...
+        {
+          errorSummaryLinkedField: 'session-details-end-time-hour',
+          formFields: ['session-details-end-time-hour'],
+          message: EditSessionController.durationLongerThanOriginallyScheduledErrorMessage,
+        },
+      ],
+    }
+  }
+
+  private static durationInMinutes(time: { hour: number; minutes: number; amOrPm: 'AM' | 'PM' }): number {
```
raby, line 124: `EditSessionController` has its own private `durationInMinutes` (line 123) and `EditSessionDateAndTimeFormForm` defines the same calculation as an inner function inside `createSessionDetailsValidations()`. If the AM/PM conversion logic ever changes, it'd need updating in both places. Extract to DateUtils?

`> // In DateUtils:
> static durationInMinutes(
>   startHour: number, startMin: number, startPeriod: 'AM' | 'PM',
>   endHour: number, endMin: number, endPeriod: 'AM' | 'PM'
> ): number {
>   const start = DateUtils.convertTo24Hour(startHour, startMin, startPeriod)
>   const end = DateUtils.convertTo24Hour(endHour, endMin, endPeriod)
>   return end.hour * 60 + end.minute - (start.hour * 60 + start.minute)
> }`

`//an idea for in the controller to use dateutils happy to change
private static isSubmittedDurationShorterThanCurrent(...): boolean {
  const currentDuration = DateUtils.durationInMinutes(...)
  const submittedDuration = DateUtils.durationInMinutes(...)
  return submittedDuration < currentDuration
}`

server/editSession/dateAndTime/editSessionDateAndTimeForm.ts
```diff
@@ -57,14 +64,18 @@ export default class EditSessionDateAndTimeFormForm {
...
-    function convertTo24Hour(hour: number, minute: number, period: 'AM' | 'PM'): { hour: number; minute: number } {
-      let hour24 = hour
-      if (period === 'AM') {
-        hour24 = hour === 12 ? 0 : hour // 12 AM is midnight (0), other AM hours stay the same
-      } else {
-        hour24 = hour === 12 ? 12 : hour + 12 // 12 PM stays 12, other PM hours add 12
-      }
-      return { hour: hour24, minute }
+    const { isSessionEnded } = this
+    function durationInMinutes(
```
raby: The inner functions `durationInMinutes`, `isTimeBefore`, `startTimeChanged`, and `endTimeChanged` are defined inside `createSessionDetailsValidations()`. This means they're re-created on every call and can't be tested in isolation. Since DateUtils is already imported, it would be cleaner to move these to that utility class. Nice-to-have, not a blocker. This keeps `createSessionDetailsValidations()` focused on wiring up validators, not computing durations :-) 
 

Example:

`// Instead of defining durationInMinutes inside createSessionDetailsValidations(),
// move it to DateUtils and call it from there:

// In DateUtils:
static durationInMinutes(
  startHour: number, startMinute: number, startPeriod: 'AM' | 'PM',
  endHour: number, endMinute: number, endPeriod: 'AM' | 'PM'
): number {
  const start = DateUtils.convertTo24Hour(startHour, startMinute, startPeriod)
  const end = DateUtils.convertTo24Hour(endHour, endMinute, endPeriod)
  return end.hour * 60 + end.minute - (start.hour * 60 + start.minute)
}

// Then in the form:
const duration = DateUtils.durationInMinutes(startHour, startMinute, startPeriod, endHour, endMinute, endPeriod)`

server/editSession/editSessionController.ts
Alison-Weatherston: Comments addressed @Raby - thank you

server/editSession/dateAndTime/editSessionDateAndTimeForm.ts
Alison-Weatherston: Comments addressed @Raby - thank you

server/editSession/editSessionController.ts
raby: Thank you

server/editSession/dateAndTime/editSessionDateAndTimeForm.ts
raby: Thank you

raby (approved):
Great additional tests

PR #586, opened 2026-06-11
No ticket/remove convert to title case

niaevans9:
Removing convert to title case from the service user banner as it was causing inconsistencies with how a person name is displayed on the service banner vs the rest of the page when the name contains multiple capitals

Before
Screenshot 2026-06-10 at 16 33 47

After
Screenshot 2026-06-11 at 07 24 23


Alison-Weatherston (approved):
LGTM

PR #588, opened 2026-06-11
Correcting cancel link

niaevans9:
When removing someone from a group the cancel button should return them to the allocated list for that group


raby (approved):
LGTM

PR #601, opened 2026-06-19
session timeout modal

Alison-Weatherston:
'server/middleware/setUpAuthentication.ts' and 'assets/js/index.js' - help from LLM for most of the code.

For testing purposes, these 2 lines in the 'assets/index.js' file can be changed so you don't need to wait 50 mins then 10 mins to test.

```
 const INACTIVITY_TIMEOUT_MS = 50 * 60 * 1000
 const MODAL_RESPONSE_TIMEOUT_MS = 10 * 60 * 1000
```

Amend to something like this:
```
 const INACTIVITY_TIMEOUT_MS = 0.1 * 60 * 1000
 const MODAL_RESPONSE_TIMEOUT_MS = 0.1 * 60 * 1000
```

Modal code in 

- assets/scss/components/_modalTimeOut.scss
- assets/js/modelTimeOutDialog.js
- server/views/partials/layout.njk

is taken from MOJ design components: https://design-patterns.service.justice.gov.uk/components/timeout-warning/#code-tab

After 50 mins this modal will be displayed:

Screenshot 2026-06-19 at 10 44 29

Then, after a further 10 mins, the user is logged out.

When the user logs back in, they will be taken to the page they were on before they were logged out.

### DIFF
--- a/assets/js/index.js
+++ b/assets/js/index.js
@@ -7,8 +7,8 @@ mojFrontend.initAll()
 
 const $inactivityWarningModal = document.querySelector('[data-modal-type="inactivity-warning"]')
 if ($inactivityWarningModal) {
-  const INACTIVITY_TIMEOUT_MS = 50 * 60 * 1000
-  const MODAL_RESPONSE_TIMEOUT_MS = 10 * 60 * 1000
+  const INACTIVITY_TIMEOUT = 0.1 * 60 * 1000
+  const MODAL_LOGOUT_TIMEOUT = 0.2 * 60 * 1000
   const SIGN_OUT_PATH = '/sign-out'
   const activityEvents = ['click', 'keydown', 'mousemove', 'scroll', 'touchstart']
   let inactivityTimerId
@@ -27,7 +27,7 @@ if ($inactivityWarningModal) {
 
   const startSignOutTimer = () => {
     clearSignOutTimer()
-    signOutTimerId = window.setTimeout(redirectToSignOut, MODAL_RESPONSE_TIMEOUT_MS)
+    signOutTimerId = window.setTimeout(redirectToSignOut, MODAL_LOGOUT_TIMEOUT)
   }
 
   const startInactivityTimer = () => {
@@ -36,7 +36,7 @@ if ($inactivityWarningModal) {
       if (!modalDialog.isOpen()) {
         modalDialog.open()
       }
-    }, INACTIVITY_TIMEOUT_MS)
+    }, INACTIVITY_TIMEOUT)
   }
 
   const handleActivity = () => {
--- a/assets/js/index.js
+++ b/assets/js/index.js
@@ -53,11 +53,11 @@ if ($inactivityWarningModal) {
     },
   })
 
-  if (!modalDialog) return
+  if (modalDialog) {
+    activityEvents.forEach(eventName => {
+      document.addEventListener(eventName, handleActivity, { passive: true })
+    })
 
-  activityEvents.forEach(eventName => {
-    document.addEventListener(eventName, handleActivity, { passive: true })
-  })
-
-  startInactivityTimer()
+    startInactivityTimer()
+  }
 }
--- a/assets/js/index.js
+++ b/assets/js/index.js
@@ -1,5 +1,61 @@
 import * as govukFrontend from 'govuk-frontend'
 import * as mojFrontend from '@ministryofjustice/frontend'
+import ModalDialog from './modelTimeOutDialog.js'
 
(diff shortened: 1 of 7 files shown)


assets/js/index.js
```diff
@@ -1,5 +1,61 @@
 import * as govukFrontend from 'govuk-frontend'
 import * as mojFrontend from '@ministryofjustice/frontend'
+import ModalDialog from './modelTimeOutDialog.js'
 
 govukFrontend.initAll()
 mojFrontend.initAll()
+
+const $inactivityWarningModal = document.querySelector('[data-modal-type="inactivity-warning"]')
+if ($inactivityWarningModal) {
+  const INACTIVITY_TIMEOUT = 0.1 * 60 * 1000
```
raby: layout.njk lines 71–72 says 50 mins and 10 mins. Should it be?

`const INACTIVITY_TIMEOUT = 50 * 60 * 1000
const MODAL_LOGOUT_TIMEOUT = 10 * 60 * 1000`

assets/js/index.js
```diff
@@ -1,5 +1,61 @@
...
+
+  const startSignOutTimer = () => {
+    clearSignOutTimer()
+    signOutTimerId = window.setTimeout(redirectToSignOut, MODAL_LOGOUT_TIMEOUT)
+  }
+
+  const startInactivityTimer = () => {
+    window.clearTimeout(inactivityTimerId)
+    inactivityTimerId = window.setTimeout(() => {
+      if (!modalDialog.isOpen()) {
```
raby, line 36: modelTimeOutDialog.js line 62 returns undefined when dialog is not supported. If this edge case happens, the following line 36 and 43 calls throw a TypeError.

`modalDialog = new ModalDialog($inactivityWarningModal).init({ ... })
if (!modalDialog) return`

server/views/partials/layout.njk
```diff
@@ -53,9 +54,35 @@
...
 
 {% block bodyStart %}
+<div class="govuk-modal-dialogue-inert-container">
 {% endblock %}
 
 {% block bodyEnd %}
+</div>
+<div class="govuk-modal-dialogue" data-modal-type="inactivity-warning" data-inert-container=".govuk-modal-dialogue-inert-container">
+  <div class="govuk-modal-dialogue__wrapper">
+    <dialog class="govuk-modal-dialogue__box" aria-labelledby="modal-title-inactivity-warning" aria-modal="true" role="modal" tabindex="-1">
```
raby: `modal` may not a recognised ARIA role and will cause accessibility failures. The correct value is `role="dialog"`??  aria-modal="true" is correctly set. (got this from an llm, ignore if it is muddled!)

assets/scss/components/_modalTimeOut.scss
```diff
@@ -0,0 +1,155 @@
...
+  }
+
+  // New dialogue width, inline button + link
+  @include govuk-media-query($from: tablet) {
+    .govuk-modal-dialogue__content {
+      padding-top: 0;
+    }
+
+    .govuk-modal-dialogue__box {
+      width: percentage($govuk-dialogue-width / map-get($govuk-breakpoints, desktop));
```
raby: This project uses Dart Sass with `@use` syntax (see line 1 of this file and all other SCSS files).

The `/` operator for division is removed in Dart Sass 2.0, this will become a compilation error on the next major Sass upgrade. Could replace with:

`@use 'sass:math';
// ...
width: percentage(math.div($govuk-dialogue-width, map-get($govuk-breakpoints, desktop)));`

server/views/partials/layout.njk
Alison-Weatherston: This is taken from the MOJ component code. However, on a little further investigation, I have changed it to role="alertdialog" because, for our particular user case, this is more appropriate because the modal requires immediate user action.

assets/js/index.js
Alison-Weatherston: I have made a little amendment

assets/js/index.js
Alison-Weatherston: Well spotted! I changed this as I was trying to make it easier for the design team to test. However, I think this is rather a dangerous way to go about it. So i have changed it to what it's meant to be.

assets/scss/components/_modalTimeOut.scss
Alison-Weatherston: Again, this is taken from the MOJ component code. I have updated it, though as it does show as an error on our pipeline too otherwise.

raby (approved):

PR #602, opened 2026-06-19
400, 500, 503 errors

Alison-Weatherston:
The 400 error can be tested locally. 

We are going to have a discussion on the 503 error. Design wants this page to have more information on why it's down and the date/time it is expected to be operational. For now though, it just has the heading text

### DIFF
--- a/server/errorHandler.test.ts
+++ b/server/errorHandler.test.ts
@@ -13,14 +13,17 @@ afterEach(() => {
 })
 
 describe('GET 404', () => {
-  it('should render content with stack in dev mode', () => {
+  it('should render content without stack in dev mode', () => {
     return request(app)
       .get('/unknown')
       .expect(404)
       .expect('Content-Type', /html/)
       .expect(res => {
-        expect(res.text).toContain('NotFoundError: Not Found')
-        expect(res.text).not.toContain('Something went wrong. The error has been logged. Please try again')
+        expect(res.text).toContain('Page not found')
+        expect(res.text).toContain('If you typed the web address, check it is correct.')
+        expect(res.text).toContain('Case list')
+        expect(res.text).not.toContain('NotFoundError: Not Found')
+        expect(res.text).not.toContain('Not found')
       })
   })
 
@@ -30,7 +33,9 @@ describe('GET 404', () => {
       .expect(404)
       .expect('Content-Type', /html/)
       .expect(res => {
-        expect(res.text).toContain('Something went wrong. The error has been logged. Please try again')
+        expect(res.text).toContain('Page not found')
+        expect(res.text).toContain('If you typed the web address, check it is correct.')
+        expect(res.text).toContain('Case list')
         expect(res.text).not.toContain('NotFoundError: Not Found')
       })
   })
--- a/server/errorHandler.test.ts
+++ b/server/errorHandler.test.ts
@@ -13,7 +13,7 @@ afterEach(() => {
 })
 
 describe('GET 404', () => {
-  it('should render content without stack in dev mode', () => {
+  it('should render 404 content without debug details', () => {
     return request(app)
       .get('/unknown')
       .expect(404)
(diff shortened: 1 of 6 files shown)


server/errorHandler.test.ts
```diff
@@ -30,7 +33,9 @@ describe('GET 404', () => {
       .expect(404)
       .expect('Content-Type', /html/)
       .expect(res => {
-        expect(res.text).toContain('Something went wrong. The error has been logged. Please try again')
+        expect(res.text).toContain('Page not found')
+        expect(res.text).toContain('If you typed the web address, check it is correct.')
+        expect(res.text).toContain('Case list')
         expect(res.text).not.toContain('NotFoundError: Not Found')
       })
```
raby, line 64: The two tests look identical in what they assert. That's because 404 is always hidden from debug output regardless of the production flag so the flag never actually gets tested here. The place `production` matters is a status like 500, where dev mode should show the stack and prod mode shouldn't. Without this, the `!production` check in `errorHandler.ts` is untested.. Worth adding something similar to:

LLM
> describe('GET 500', () => {
>   it('should show stack trace in dev mode', () => {
>     return request(appWithAllRoutes({ production: false }))
>       .get('/route-that-throws-500')
>       .expect(500)
>       .expect(res => {
>         expect(res.text).toContain('Sorry, there is a problem with the service')
>         expect(res.text).toContain('Error:') // stack visible in dev
>       })
>   })
>
>   it('should NOT show stack trace in production mode', () => {
>     return request(appWithAllRoutes({ production: true }))
>       .get('/route-that-throws-500')
>       .expect(500)
>       .expect(res => {
>         expect(res.text).toContain('Sorry, there is a problem with the service')
>         expect(res.text).not.toContain('Error:') // stack hidden in prod
>       })
>   })
> })

server/errorHandler.test.ts
```diff
@@ -13,14 +13,17 @@ afterEach(() => {
 })
```
raby, line 14: There's no test coverage for the 500 or 503 content added to `errorPageContentByStatus`. If someone accidentally changes the heading string for 500, nothing would catch it. We could add:

LLM:
`> describe('GET 503', () => {
>   it('should render the 503 heading', () => {
>     // simulate a 503 from a downstream service
>     return request(app)
>       .get('/route-that-returns-503')
>       .expect(503)
>       .expect(res => {
>         expect(res.text).toContain('Sorry, the service is unavailable')
>       })
>   })
> })`

server/errorHandler.ts
```diff
@@ -1,6 +1,35 @@
...
+    heading: 'Page not found',
+    body: '<p>If you typed the web address, check it is correct.</p><p>If you pasted the web address, check you copied the entire address.</p>',
+  },
+  500: {
+    heading: 'Sorry, there is a problem with the service',
+    body: '<p>Try reloading the page. To do this, press F5 (on a PC), or Cmd + R (on a Mac).</p><p>If the page still does not load, try again later.</p><p>You can report a problem with the service through the <a href="https://mojprod.service-now.com/moj_sp?id=emp_taxonomy_topic&topic_id=75608d7847d099106322862c736d43ac">Tech Portal.</a></p>',
+  },
+  503: {
+    heading: 'Sorry, the service is unavailable',
+    body: '',
```
raby, line 22: An empty body renders an empty div? Could add <p>Try again later.</p>? Probably a design decision though out of scope.

Alison-Weatherston: This is here until we have a chat about it and get some sort of decision on what the text should say.

server/errorHandler.test.ts
Alison-Weatherston: I have fixed this issue

server/errorHandler.test.ts
Alison-Weatherston: Tests added

server/errorHandler.ts
raby: Thanks

raby (approved):
Looks good

PR #608, opened 2026-06-23
amending the ALLOWED_PAST_DATE_REGIONS to N54

Alison-Weatherston:
We want the North East (N54) to be the only group able to create groups in the past. Previously this was restricted to the the East Midlands (N53).


DanEgan1 (approved):

raby (commented):
LGTM

PR #610, opened 2026-06-24
APG-1956: duplication removed on status history

Alison-Weatherston:
Figma example\:| https://www.figma.com/design/Y1h09XHyEI9O8KZoOpvlTS/Accredited-Programmes--Community-?node-id=12903-64684&t=2uoSP8pYG3OnuLID-1

'Status update' text added to each line instead of the status which is currently showing

After:

Screenshot 2026-06-24 at 14 21 01

Before:

Screenshot 2026-06-24 at 14 22 06


raby (approved):
Looks good

PR #612, opened 2026-06-24
Apg 2350/block unapproved regions from homepage

DanEgan1:
The new page
image

### DIFF
--- a/server/home/homeController.test.ts
+++ b/server/home/homeController.test.ts
@@ -20,10 +20,13 @@ describe('HomeController', () => {
       }
       app = TestUtils.createTestAppWithSession(sessionData, {})
 
-      return request(app).get('/').expect(200).expect(res => {
-        expect(res.text).toContain('Accredited Programmes')
-        expect(res.text).not.toContain('You cannot access Accredited Programmes services yet')
-      })
+      return request(app)
+        .get('/')
+        .expect(200)
+        .expect(res => {
+          expect(res.text).toContain('Accredited Programmes')
+          expect(res.text).not.toContain('You cannot access Accredited Programmes services yet')
+        })
     })
 
     it('renders the invalid region page when region is not allowed', async () => {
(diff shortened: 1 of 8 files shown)


server/home/homeController.ts
```diff
@@ -15,6 +16,11 @@ export default class HomeController extends BaseController {
     const presenter = new HomePresenter()
     const view = new HomeView(presenter)
 
+    if (!isRegionAllowed(req.session.userRegion?.regionCode)) {
+      res.status(401)
```
DanEgan1: Feels like a unauthorised repsonse code is appropriate.

raby: 401 Unauthorized means "you need to authenticate". 403 Forbidden means "you are authenticated but not permitted". The user IS authenticated , their region just isn't approved yet. This could be a 403? Happy with 401 though, could comment if 401: 

`// 401 is intentional here — users from unapproved regions are treated as
// unauthorised to use this service at this stage of rollout`

server/home/homeController.test.ts
```diff
@@ -0,0 +1,82 @@
...
+      app = TestUtils.createTestAppWithSession(sessionData, {})
+
+      return request(app)
+        .get('/')
+        .expect(200)
+        .expect(res => {
+          expect(res.text).toContain('Accredited Programmes')
+          expect(res.text).not.toContain('You cannot access Accredited Programmes services yet')
+        })
+    })
```
raby, line 94: There is no test for the case where `req.session.userRegion `is completely absent (e.g. `setUpUserLocation` middleware didn't run, or session is fresh). `isRegionAllowed(undefined)` returns `false` when restriction is enabled, so those users would see the invalid region page. This behaviour should be explicitly tested as a user with no region yet may need a different error message.

`it('renders the invalid region page when userRegion is undefined', async () => {
  const sessionData: Partial<SessionData> = {
    userRegion: undefined,
  }
  app = TestUtils.createTestAppWithSession(sessionData, {})

  return request(app)
    .get('/')
    .expect(403) // or 401 if kept — update once status code is fixed
    .expect(res => {
      expect(res.text).toContain('You cannot access Accredited Programmes services yet')
    })
})`

server/home/homeController.ts
DanEgan1: Makes sense good point

raby (dismissed):
Looks good

raby (approved):

PR #614, opened 2026-06-24
APG-2390 Fixed issue not allowing auto reschedule of past sessions being moved into the future

Jack-Dempster:


niaevans9 (approved):
lgtm

PR #617, opened 2026-06-25
Get the referral person name from the API instead of the compound ID on the radio button

niaevans9:
Get the referral person name from the API instead of the compound ID on the radio button


raby (approved):
Looks good

PR #620, opened 2026-06-25
Downgrade ESLint

niaevans9:
Downgrade of ESLint as v10 isnt compatible with eslint-plugin-import@2.32.0 and its causing issues
https://github.com/import-js/eslint-plugin-import/issues/3227


Alison-Weatherston (approved):
LGTM

PR #626, opened 2026-06-26
Empty table heading in allocations and waitlist

Alison-Weatherston:
Table headings 'Add to group' and 'Remove from group' added.

Screenshot 2026-06-26 at 10 17 39

Screenshot 2026-06-26 at 10 17 57


raby (approved):
Looks good

PR #628, opened 2026-06-26
handle dev regions

DanEgan1:
Due to Dev users having different region codes to prod delius need to check a different list of region codes

### DIFF
--- a/server/home/allowedRegions.ts
+++ b/server/home/allowedRegions.ts
@@ -1,11 +1,20 @@
 import config from "../config"
 
 /**
- * For info: N53 is the east midlands, N54 is the North East.
+ * For info:
+ * N53 Is the East Midlands
+ * N54 Is the North East
  * Full region code mapping can be found at:
  * https://dsdmoj.atlassian.net/wiki/spaces/IC/pages/6027477185/Interventions+Manager+Data+Migration+Strategy#2.-Phased-out-Data-Migration
  */
 export const ALLOWED_REGIONS: string[] = ['N53', 'N54']
+
+/**
+ * For info:
+ * N02 Is an alternate region code for the north east used in test Delius accounts, but not used in production.
+ * N07 Is the region code for London. Currently used for test accounts but not yet allowed in production.
+ * Other are descibed in ALLOWED_REGIONS above.
+ */
 export const DEV_ALLOWED_REGIONS: string[] = ['N01', 'N07', 'N53', 'N54'] // Dev users do not have access to the same regions in delius so extra ones allowed.
 
 /**
--- a/server/home/allowedRegions.ts
+++ b/server/home/allowedRegions.ts
@@ -1,3 +1,5 @@
+import config from "../config"
+
 /**
  * For info: N53 is the east midlands, N54 is the North East.
  * Full region code mapping can be found at:
@@ -12,11 +14,11 @@ export const DEV_ALLOWED_REGIONS: string[] = ['N01', 'N07', 'N53', 'N54'] // Dev
  * When set to 'false', all regions can access the service.
  */
 function isRegionRestrictionEnabled(): boolean {
-  return process.env.ENABLE_REGION_RESTRICTION === 'true'
+  return config.enable_region_restriction === true
 }
 
 function getAllowedRegions(): string[] {
-  return process.env.ENVIRONMENT_NAME === 'DEV' ? DEV_ALLOWED_REGIONS : ALLOWED_REGIONS
+  return config.environmentName?.toUpperCase()  === 'DEV' ? DEV_ALLOWED_REGIONS : ALLOWED_REGIONS
 }
 
 export function isRegionAllowed(regionCode: string | undefined): boolean {
--- a/server/home/allowedRegions.ts
+++ b/server/home/allowedRegions.ts
@@ -1,4 +1,4 @@
-import config from "../config"
+import config from '../config'
 
 /**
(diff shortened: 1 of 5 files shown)


server/home/allowedRegions.ts
```diff
@@ -4,6 +4,7 @@
  * https://dsdmoj.atlassian.net/wiki/spaces/IC/pages/6027477185/Interventions+Manager+Data+Migration+Strategy#2.-Phased-out-Data-Migration
  */
 export const ALLOWED_REGIONS: string[] = ['N53', 'N54']
+export const DEV_ALLOWED_REGIONS: string[] = ['N01', 'N07', 'N53', 'N54'] // Dev users do not have access to the same regions in delius so extra ones allowed.
```
DanEgan1, line 18: These are the region codes used in dev for the past month or so

N02:10004 - North east
N07:2305 - London This is a real region code but used by real accounts used for testing
wiremocked_region_code:723 Local code an not used anymore
N56:7 Decided not to include this as its a real region code and used very little looked to be Nias account associated with it better handled by turning of flag for dev once tested I think

server/home/allowedRegions.ts
```diff
@@ -14,13 +15,16 @@ function isRegionRestrictionEnabled(): boolean {
   return process.env.ENABLE_REGION_RESTRICTION === 'true'
 }
 
+function getAllowedRegions(): string[] {
```
raby, line 29: The `helm_deploy/values-dev.yaml` correctly uses `ENVIRONMENT_NAME: DEV` (uppercase), so the deployed dev environment works fine. But a local developer using `feature.env` and turning `ENABLE_REGION_RESTRICTION=true` on locally would find the `DEV` list is silently ignored, they may see N01 rejected and have no obvious reason why. 

Could make case insensitive here or update Or update `feature.env` to match: `ENVIRONMENT_NAME=DEV`

`// option A: case-insensitive compare
function getAllowedRegions(): string[] {
  return process.env.ENVIRONMENT_NAME?.toUpperCase() === 'DEV'
    ? DEV_ALLOWED_REGIONS
    : ALLOWED_REGIONS
}`

server/home/allowedRegions.ts
```diff
@@ -14,13 +15,16 @@ function isRegionRestrictionEnabled(): boolean {
...
+  return process.env.ENVIRONMENT_NAME === 'DEV' ? DEV_ALLOWED_REGIONS : ALLOWED_REGIONS
```
raby: `process.env `read directly instead of via `config.ts`?

config.ts already has: `environmentName: get('ENVIRONMENT_NAME', ''),`

But` allowedRegions.ts` reads `process.env.ENVIRONMENT_NAME` directly. This is inconsistent,  if someone renamed the env var in config.ts they might miss this? The pattern used elsewhere in the app is to go through the central config.

Suggested improvement:
`import config from '../config'

function getAllowedRegions(): string[] {
  return config.environmentName?.toUpperCase() === 'DEV'
    ? DEV_ALLOWED_REGIONS
    : ALLOWED_REGIONS
}`

Same applies to` isRegionRestrictionEnabled() `which also reads `process.env` directly for `ENABLE_REGION_RESTRICTION` — worth noting as a pre-existing pattern following, but this PR is a good opportunity to tidy it?

server/home/allowedRegions.ts
raby: The comment explains the reasoning which is good however may be worth a test or an inline code comment to make this explicit otherwise a future dev will look at the data and wonder "why isn't N02 in here?"

LLM:
`// N02 is a real prod region code. Although it appears in dev data (10004 accounts),
// it is intentionally excluded to avoid accidentally granting access if N02 is
// ever rolled out to prod. See APG-2350 PR discussion.
export const DEV_ALLOWED_REGIONS: string[] = ['N01', 'N07', 'N53', 'N54']`

server/home/homeController.test.ts
```diff
@@ -108,4 +122,44 @@ describe('HomeController', () => {
...
+      app = TestUtils.createTestAppWithSession(sessionData, {})
+
+      return request(app)
+        .get('/')
+        .expect(403)
+        .expect(res => {
+          expect(res.text).toContain('You cannot access the Accredited Programmes service yet')
+        })
+    })
+  })
```
raby, line 164: Currently `allowedRegions.ts` is tested indirectly through the controller integration tests. That's valid, but a dedicated unit test (e.g. `allowedRegions.test.ts`) would make the logic easier to test in isolation. especially as the region lists grow. Example:

`// allowedRegions.test.ts
import { isRegionAllowed } from './allowedRegions'

describe('isRegionAllowed', () => {
  beforeEach(() => {
    process.env.ENABLE_REGION_RESTRICTION = 'true'
  })
  afterEach(() => {
    delete process.env.ENABLE_REGION_RESTRICTION
    delete process.env.ENVIRONMENT_NAME
  })

  it('returns true for N53 in any environment', () => {
    expect(isRegionAllowed('N53')).toBe(true)
  })

  it('returns false for N01 in non-DEV', () => {
    expect(isRegionAllowed('N01')).toBe(false)
  })

  it('returns true for N01 when ENVIRONMENT_NAME=DEV', () => {
    process.env.ENVIRONMENT_NAME = 'DEV'
    expect(isRegionAllowed('N01')).toBe(true)
  })
})`

raby (approved):
Looks good

PR #629, opened 2026-06-29
Update CSS to allow filter to appear above results on zooming

niaevans9:
Caselist and Group filters overlap the content when zoomed in
Fix to stack the filter over the results when the user is zoomed in or the viewport is small

Before
image
After
image

### DIFF
--- a/assets/scss/overrides/_local.scss
+++ b/assets/scss/overrides/_local.scss
@@ -47,6 +47,39 @@
     width: 100%;
   }
 }
+
+// Responsive filter layout for accessibility
+.moj-filter {
+  max-width: 100%;
+}
+
+.caselist-filter-column,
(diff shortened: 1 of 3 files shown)


assets/scss/overrides/_local.scss
```diff
@@ -47,6 +47,39 @@
...
+  min-width: 0;
+}
+
+// Stack filter and results vertically when viewport is narrow
+// Allow users to see full filter options when zoomed in
+@media (max-width: 64em) {
+  .caselist-filter-column,
+  .caselist-results-column,
+  .group-filter-column,
+  .group-results-column{
```
raby, line 74: nit: inconsistent missing space may trigger a linter somewhere

raby (approved):
Looks good

PR #630, opened 2026-06-29
Improve hidden label to be more meaningful to screen readers

niaevans9:
Improve the hidden text shown for screenreaders when selecting a check box on the attendance and session notes page

Screenshot 2026-06-29 at 11 28 32
 

Before
``` html
<label for="person-1">
      <span class="govuk-visually-hidden">
        Select [object Object]
      </span>
 </label>
```
After
``` html
<label class="govuk-label govuk-checkboxes__label" for="multi-select-attendance-multi-select-row-1">
            <span class="govuk-visually-hidden">Select Aisha McGlynn</span>
          </label>
```


raby (approved):
LGTM

PR #640, opened 2026-07-01
Add initial audit to test setup

DanEgan1:
Only one audit included so it can be tested in dev. Went with LDC as its straight forward to test.

In local testing recieved exception (Expected behaviour) which was then handled and logged. This didnt cause a changein behaviour otherwise so shouldnt disrupt dev. 
Change is flagged off in PROD and Preprod


server/data/index.ts
```diff
@@ -35,11 +34,10 @@ export const dataAccess = () => {
   return {
     applicationInfo,
     hmppsAuthClient,
-    hmppsAuditClient: new HmppsAuditClient(config.sqs.audit),
```
DanEgan1, line 38: Audit client no longer used. It was used i old audit service which wasnt used anywhere and has been replaced

server/services/auditService.test.ts
```diff
@@ -1,57 +1,69 @@
-import AuditService, { Page } from './auditService'
-import HmppsAuditClient from '../data/hmppsAuditClient'
+import { auditService } from '@ministryofjustice/hmpps-audit-client'
```
DanEgan1, line 1: Files are pretty much new might be easier to review outside of comparison

server/ldc/ldcController.ts
```diff
@@ -25,6 +26,10 @@ export default class LdcController extends BaseController {
     )
 
     if (req.method === 'POST') {
+      sendAuditEvent('EDIT_REFFERAL_LDC', username, referralDetails?.crn, 'CRN', {
```
DanEgan1: CRN used to stick to subject type guidlines
https://dsdmoj.atlassian.net/wiki/spaces/HAA/pages/4546527310/HMPPS+Audit+-+Subject+Type+Register?xpis=eyJicmlkZ2UiOiJzbWFydExpbmtzIiwiaWQiOiIxNzgyOTEzMTY2MjkxIiwic291cmNlIjoiY29uZmx1ZW5jZSJ9

helm_deploy/values-dev.yaml
```diff
@@ -15,7 +15,7 @@ generic-service:
     EXAMPLE_API_URL: 'https://template-kotlin-dev.hmpps.service.justice.gov.uk'
     ACCREDITED_PROGRAMMES_MANAGE_AND_DELIVER_URL: 'https://accredited-programmes-manage-and-deliver-api-dev.hmpps.service.justice.gov.uk'
     ENVIRONMENT_NAME: DEV
-    AUDIT_ENABLED: "false"
+    AUDIT_ENABLED: "true"
```
DanEgan1, line 18: Still set to false in prod and preprod

server/services/auditService.ts
```diff
@@ -1,29 +1,31 @@
...
-  async logAuditEvent(event: AuditEvent) {
-    await this.hmppsAuditClient.sendMessage(event)
+export default async function sendAuditEvent(
+  action: string,
+  username: string,
+  subjectId?: string,
+  subjectType?: string,
+  details?: object,
+) {
+  // Check if audit is enabled for enviroment
```
raby: Nit: Typo: enviroment → environment

server/ldc/ldcController.ts
raby: EDIT_REFERRAL_LDC (accidental double F)

server/services/auditService.test.ts
```diff
@@ -1,57 +1,69 @@
...
+
+      expect(auditService.sendAuditMessage).not.toHaveBeenCalled()
+    })
+
+    it('should send audit message when audit is enabled', async () => {
+      ;(config as jest.Mocked<typeof config>).sqs.audit.enabled = true
+      ;(auditService.sendAuditMessage as jest.Mock).mockResolvedValue(undefined)
+
+      await sendAuditEvent('EDIT_REFFERAL_LDC', 'testuser123', 'subject123', 'CRN', {
+        referralId: 'refferalId',
```
raby, line 29: referralId double f accident

server/ldc/ldcController.ts
raby: `sendAuditEvent(...) `is called without await. The audit function handles its own errors internally so it won't crash, but the redirect can happen before the message is sent to SQS. If fire-and-forget is intentional (to avoid blocking the user journey on SQS latency), we could make that explicit with void:

llm:

`> // If intentional fire-and-forget:
> void sendAuditEvent('EDIT_REFERRAL_LDC', username, referralDetails?.crn, 'CRN', {
>   referralId,
>   hasLdc: req.body.hasLdc,
> })
>
> // Or if audit should complete before redirecting:
> await sendAuditEvent('EDIT_REFERRAL_LDC', username, referralDetails?.crn, 'CRN', {
>   referralId,
>   hasLdc: req.body.hasLdc,
> })`

DanEgan1: Had a bit of a look at how other places handle it. Most seemed to use an await though some didnt. So gone with that. Im a bit 50/50 on which ways best but await seems to align with whats suggested in the audit readme so gone with that.

DanEgan1: Checked and its not taking longer in testing but ill double check when its deployed to dev

raby: Thank you - I had split thoughts too - well done for the extra research

raby (approved):
Looks good

PR #645, opened 2026-07-02
Treatment managers and facilitators duplication errors fix

Alison-Weatherston:
Error messages are now showing correctly above each field that has a duplicate.

Screenshot 2026-07-03 at 08 57 36


raby (approved):
Looks great!

PR #648, opened 2026-07-03
Plural or singular text in results

Alison-Weatherston:
Shows the word 'result' instead of 'results' when there is only 1 result.

Screenshot 2026-07-03 at 09 10 51

### DIFF
--- a/server/caselist/caselistPresenter.ts
+++ b/server/caselist/caselistPresenter.ts
@@ -247,13 +247,17 @@ export default class CaselistPresenter {
   }
 
   generateNoResultsString(): string {
+    const count = this.otherCaselistCountTotal
+    const resultsText = `${count} ${count === 1 ? 'result' : 'results'}`
+
     if (this.section === CaselistPageSection.Open) {
-      return this.otherCaselistCountTotal === 0
+      return count === 0
         ? 'No results found. Clear or change the filters.'
-        : `No results in open referrals. ${this.otherCaselistCountTotal} results in closed referrals.`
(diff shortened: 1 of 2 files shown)


server/caselist/caselistPresenter.ts
```diff
@@ -247,13 +247,17 @@ export default class CaselistPresenter {
   }
 
   generateNoResultsString(): string {
+    const count = this.otherCaselistCountTotal
+    const resultsText = `${count} ${count === 1 ? 'result' : 'results'}`
```
raby, line 251: The local variable `resultsText` shadows the existing class getter get `resultsText()` (line 58) which formats the pagination string. TypeScript won't error here,  the local const wins inside the function ... however for readability it could confuse someone reading the method and thinking it refers to the getter. 

Suggest renaming to something like `otherResultsText`:

`> const count = this.otherCaselistCountTotal
> const otherResultsText = `${count} ${count === 1 ? 'result' : 'results'}`
>
> if (this.section === CaselistPageSection.Open) {
>   return count === 0
>     ? 'No results found. Clear or change the filters.'
>     : `No results in open referrals. ${otherResultsText} in closed referrals.`
> }
>
> return count === 0
>   ? 'No results found. Clear or change the filters.'
>   : `No results in closed referrals. ${otherResultsText} in open referrals.``

`otherResultsText` was an LLM suggestion - maybe a human brain can be more inventive :-)

raby (approved):
Looks good overall, one small comment

PR #659, opened 2026-07-07
check your answers page change destination

Alison-Weatherston:
Change links take users to the correct page but then make them step through the whole journey again rather than sending them back to the review page after they've edited that page. Should work that eg, you click the change link next to cohort, it sends you to the cohort page to edit, and then back to the review page when you click ‘Continue’.


raby (approved):
Looks good

PR #666, opened 2026-07-09
Replace drop down with checkboxes

niaevans9:
Replace the PDU dropdown on the caselist filter with checkboxes to allow multiple selections

image

### DIFF
--- a/server/caselist/caselistFilter.test.ts
+++ b/server/caselist/caselistFilter.test.ts
@@ -9,7 +9,7 @@ describe(CaselistFilter, () => {
         status: 'REFERRAL_SUBMITTED',
         cohort: 'SEXUAL_OFFENCE',
         crnOrPersonName: 'Building',
-        pdu: 'PDU3',
+        pdu: ['PDU3'],
         reportingTeam: ['Team5'],
       }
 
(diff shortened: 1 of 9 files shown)


server/caselist/caselistPresenter.ts
```diff
@@ -223,19 +223,30 @@ export default class CaselistPresenter {
...
         text: pdu.pduName,
         value: pdu.pduName,
-        selected: this.filter.pdu === pdu.pduName,
+        selected: this.filter.pdu?.includes(pdu.pduName),
       }))
       .sort((a, b) => a.text.localeCompare(b.text))
     return checkboxArgs.concat(pduCheckboxArgs)
   }
 
+  generatePDUCheckboxArgs(): CheckboxesArgsItem[] {
```
raby, line 215: With the change, does `generatePduSelectArgs()` method have zero callers and could be deleted? Could be useful too

server/caselist/caselistFilter.test.ts
```diff
@@ -21,9 +21,29 @@ describe(CaselistFilter, () => {
...
+
+    it('handles multiple pdus', () => {
+      const query = {
+        pdu: ['PDU1', 'PDU2'],
+      }
+
+      const filter = CaselistFilter.fromRequest({ query } as unknown as Request)
+
+      expect(filter.pdu).toEqual(['PDU1', 'PDU2'])
+    })
```
raby, line 46: The new validation logic in `fromRequest` that strips a `reportingTeam` when it doesn't belong to the selected PDU has no unit test in caselistFilter.test.ts. The three `fromRequest` tests cover valid combinations. We could add a test like:

```
it('strips reportingTeam when it does not belong to the selected PDU', () => {
  const query = { pdu: 'PDU1', reportingTeam: 'Team5' } // Team5 belongs to PDU3, not PDU1
  const filter = CaselistFilter.fromRequest(
    { query } as unknown as Request,
    TestUtils.createCaseListFilters().locationFilters,
  )
  expect(filter.reportingTeam).toBeUndefined()
})
```

raby: The validation is exercised indirectly via the controller-level test though after checking

server/caselist/caselistPresenter.test.ts
```diff
@@ -70,7 +70,7 @@ describe(`filters`, () => {
   describe('generatePduSelect', () => {
```
raby: Small nit: 

`describe('generatePduSelect', ...) `tests `generatePDUCheckboxArgs()`. The describe block name is the old method name and copuld be updated to 'generatePDUCheckboxArgs'?

raby (approved):
Looks great!

PR #667, opened 2026-07-10
Date format updated

Alison-Weatherston:
START DATE FORMAT AMENDED

BEFORE:

Screenshot 2026-07-10 at 09 10 39

AFTER:

Screenshot 2026-07-10 at 09 09 35

### DIFF
--- a/server/createGroup/check-your-answers/createGroupCyaPresenter.test.ts
+++ b/server/createGroup/check-your-answers/createGroupCyaPresenter.test.ts
@@ -204,7 +204,7 @@ describe('CreateGroupCyaPresenter', () => {
     it('returns summary list with all group details and no team members', () => {
       const createGroupFormData: Partial<CreateGroupRequest> = {
         groupCode: 'ABC123',
-        earliestStartDate: '2026-06-01',
+        earliestStartDate: 'Monday 30 July 2026',
         createGroupSessionSlot: [],
         cohort: 'GENERAL',
         sex: 'MALE',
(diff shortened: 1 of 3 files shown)


server/createGroup/check-your-answers/createGroupCyaPresenter.ts
```diff
@@ -20,6 +20,30 @@ export default class CreateGroupCyaPresenter {
     return 'Review your group details'
   }
 
+  private formatCyaDate(): string {
```
raby, line 25: Great idea this function, and good start. Rather than a bespoke implementation, use existing utils?

```
private formatCyaDate(): string {
  const input = this.createGroupFormData.earliestStartDate
  if (!input) return ''

  const date = DateFormatUtils.parseDate(input)  // handles both DD/MM/YYYY and YYYY-MM-DD
  if (!date) return input

  const calendarDay = CalendarDay.fromComponents(date.getDate(), date.getMonth() + 1, date.getFullYear())
  return `${DateUtils.dayOfWeek(calendarDay)} ${DateUtils.formattedDate(calendarDay)}`
}
```

server/createGroup/check-your-answers/createGroupCyaPresenter.test.ts
```diff
@@ -204,7 +204,7 @@ describe('CreateGroupCyaPresenter', () => {
     it('returns summary list with all group details and no team members', () => {
       const createGroupFormData: Partial<CreateGroupRequest> = {
         groupCode: 'ABC123',
-        earliestStartDate: '2026-06-01',
+        earliestStartDate: 'Monday 30 July 2026',
```
raby: This bypasses the formatting logic, it hits the `if (!date) return input ` and passes straight through. Replace with a real DD/MM/YYYY input like '30/07/2026' so the test actually exercises the conversion?

raby (approved):

PR #668, opened 2026-07-10
Apg 2392/add audits to user actions availability audits

DanEgan1:


server/availabilityAndMotivation/locationPreferences/locationPreferencesController.ts
```diff
@@ -31,6 +32,10 @@ export default class LocationPreferencesController extends BaseController {
       username,
     )
 
+    await sendAuditEvent('VIEW_ADD_LOCATION_PREFERENCES', username, referralDetails?.crn ?? referralId, 'CRN', {
```
raby: The VIEW audit calls in all three methods of` locationPreferencesController.ts` are placed before the `req.method === 'POST' `branch/check, meaning a form submission fires both the VIEW event and the write event in the same request I think? Maybe polluting the audit trail with a spurious view event.

Affected methods:
- `showLocationPreferencesPage` — `VIEW_ADD_LOCATION_PREFERENCES` fires at line 35, POST check is at line 53

- `showAdditionalPduLocationPreferencesPage` — `VIEW_ADDITIONAL_PDU_LOCATION_PREFERENCES` fires at line 95, POST check is at line 117

- `showCannotAttendLocationsPage` — `VIEW_CANNOT_ATTEND_LOCATIONS` fires at line 156, POST check is at line 163

The VIEW audit call should only fire on GET? We could move each one into an else branch so it is skipped on POST.

The current tests don't catch this because they only assert the audit event on GET requests, could be worth adding a test that posts and asserts VIEW_* is not called alongside the write event.

server/availabilityAndMotivation/availabilityAndMotivationController.ts
raby: (apologies not in changes I picked this up now, thought could be valuable ) 

Every other audit call in this branch uses `referralDetails?.crn ?? referralId` as the fallback. Here `subjectId` will be undefined if crn is null/absent. Should it be?

`await sendAuditEvent('EDIT_REFERRAL_MOTIVATION', username, referralDetails?.crn ?? referralId, 'CRN', {`

raby, line 92: (apologies not in changes I picked this up now, thought could be valuable )

There's an inconsistency in how details is passed to `sendAuditEvent` across the controllers. `EDIT_REFERRAL_MOTIVATION` passes a raw object:

`details: data.paramsForUpdate,`

whereas `CREATE_AVAILABILITY`, `EDIT_AVAILABILITY`, and `UPDATE_REFERRAL_LOCATION_PREFERENCES` pre-stringify:

`details: JSON.stringify(data.paramsForUpdate),`

`auditService.ts` always calls `JSON.stringify(details)` itself before sending to SQS, so the availability/location events end up double-encoded (`"{\"key\":\"val\"}"`) while the motivation event is correctly nested. We could pick one convention,  the simplest fix is to remove the inner `JSON.stringify` calls in the controllers and let the service handle serialisation consistently?

server/availabilityAndMotivation/locationPreferences/locationPreferencesController.ts
DanEgan1: Ideally we send the audit before getting the data. Gone with checking if its a GET request to prevent duplciates

raby (approved):

PR #672, opened 2026-07-14
Change PDU filter to be checkboxes

niaevans9:
Update PDU filter on the Group page to allow multiple PDUs to be selected
image

### DIFF
--- a/server/groupOverview/groupListFilter.test.ts
+++ b/server/groupOverview/groupListFilter.test.ts
@@ -7,7 +7,7 @@ describe(GroupListFilter, () => {
       const query = {
         cohort: 'SEXUAL_OFFENCE',
         groupCode: 'CODE',
-        pdu: 'PDU3',
+        pdu: ['PDU3'],
         deliveryLocations: ['delivery-location-1'],
         sex: 'Male',
       }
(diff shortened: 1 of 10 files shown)


server/groupOverview/groupListFilter.test.ts
```diff
@@ -63,17 +83,17 @@ describe(GroupListFilter, () => {
...
       it('correctly sets pdu', () => {
         const filter = new GroupListFilter()
-        filter.pdu = 'PDU1'
+        filter.pdu = ['PDU1']
 
-        expect(filter.params.pdu).toEqual('PDU1')
+        expect(filter.params.pdu).toEqual(['PDU1'])
         expect(filter.params.deliveryLocations).toEqual(undefined)
       })
     })
```
raby, line 99: No test in `groupListFilter.test.ts` or `groupPresenter.test.ts` verifies the output of `paramsAsQueryParams` when pdu has multiple values. The `getSubNavArgs` tests only use `groupCode` (a plain string). Could add something like:

`it('serialises multiple PDUs as repeated query params', () => {
  const filter = GroupListFilter.empty()
  filter.pdu = ['PDU1', 'PDU2']

  expect(filter.paramsAsQueryParams).toBe('pdu=PDU1&pdu=PDU2')
})`

server/groupOverview/groupListFilter.ts
raby, line 76: paramsAsQueryParams wasn't updated when pdu was changed from string to string[]. 

URLSearchParams.append coerces an array to a comma-joined string (pdu=PDU1%2CPDU2) instead of repeated params (pdu=PDU1&pdu=PDU2), so multi-PDU selections will be silently lost when switching between the Not started / Completed sub-nav tabs.

`> Object.entries(this.params).forEach(([key, value]) => {
>   if (Array.isArray(value)) {
>     value.forEach(v => searchParams.append(key, v))
>   } else {
>     searchParams.append(key, value)
>   }
> })`

Also could be worth adding a test: filter.pdu = ['PDU1', 'PDU2'] → expect(filter.paramsAsQueryParams).toBe('pdu=PDU1&pdu=PDU2')

raby (approved):

PR #673, opened 2026-07-14
add group audits

DanEgan1:


server/createGroup/editGroupController.ts
```diff
@@ -330,9 +367,14 @@ export default class EditGroupController extends BaseController {
     const { username } = req.user
     let formError: FormValidationError | null = null
     let userInputData = null
+    let viewAuditSent = false
     req.session.originPage = req.path
 
     if (!createGroupFormData?.pduCode) {
+      if (req.method === 'GET') {
+        await sendAuditEvent('VIEW_EDIT_GROUP_PDU', username, groupId, 'NOT_APPLICABLE')
```
DanEgan1: Debatable is the best way to handle this but felt the most consistent.

Audit team say is better to audit before making the call.

server/createGroup/editGroupController.ts
```diff
@@ -358,6 +400,9 @@ export default class EditGroupController extends BaseController {
         return res.redirect(`/${groupId}/edit-group-delivery-location`)
       }
     }
+    if (!viewAuditSent) {
```
raby: The `if (!viewAuditSent)` check at line 403 (and line 466 for the location handler) has a flaw. 

`viewAuditSent` is only ever set to true inside a `req.method === 'GET'` branch on line 374. This means when a user submits the form (POST) and it fails validation (e.g. no PDU selected), there is no redirect — so execution falls through to line 403, and `VIEW_EDIT_GROUP_PDU` fires on a POST request, not a GET.

The `viewAuditSent` flag tracks whether we've sent the audit, but it doesn't know why it wasn't sent. It can't tell the difference between "we haven't sent it because this is the first GET load" vs "we haven't sent it because this is a POST with errors".

We could add a `req.method === 'GET'` guard on the fallthrough:

```
> // line 403 — change this:
> if (!viewAuditSent) {
>
> // to this:
> if (!viewAuditSent && req.method === 'GET') {
```

Apply the same fix at line 466 for editGroupLocation. The simpler handlers in this file (e.g. gender at line 209, cohort at line 262) do this correctly with a single `if (req.method === 'GET')` at the top — that pattern is cleaner if you want to refactor.

server/createGroup/editGroupController.ts
```diff
@@ -402,6 +452,9 @@ export default class EditGroupController extends BaseController {
           deliveryLocationName: data.paramsForUpdate.deliveryLocationName,
           deliveryLocationCode: data.paramsForUpdate.deliveryLocationCode,
         }
+        await sendAuditEvent('EDIT_GROUP_LOCATION', username, groupId, 'SEARCH_TERM', {
+          details: req.session.createGroupFormData,
```
raby: Line 456 sends `req.session.createGroupFormData` as the audit details. At this point in the code, after line 450–454, that object contains: `groupCode`, `pduName`, `pduCode`, `deliveryLocationCode`, and `deliveryLocationName`. The only thing that actually changed was `deliveryLocationCode` and `deliveryLocationName`. 

Compare this with every other edit audit in this file — they only log the specific field that changed:
Line 232: `{ sex: data.paramsForUpdate.sex }`
Line 285: `{ cohort: data.paramsForUpdate.cohort }`
Line 341: `{ groupCode: data.paramsForUpdate.groupCode }`

```
> await sendAuditEvent('EDIT_GROUP_LOCATION', username, groupId, 'SEARCH_TERM', {
>   details: {
>     deliveryLocationCode: data.paramsForUpdate.deliveryLocationCode,
>     deliveryLocationName: data.paramsForUpdate.deliveryLocationName,
>   },
> })
```

Note: the test assertion at `editGroupController.test.ts` uses `expect.objectContaining({ deliveryLocationCode: 'LOC-2' }) `which would pass either way — so the test won't catch this. You'll need to make it more specific too, e.g. adding a `expect.not.objectContaining({ groupCode: ... }) `or asserting the exact object shape.

server/createGroup/editGroupController.ts
raby: I could be wrong but lines 375 and 404 pass 'NOT_APPLICABLE' as the `subjectType` for `VIEW_EDIT_GROUP_PDU`, but the `subjectId` is `groupId`. Every other edit-group audit event in this file uses 'SEARCH_TERM' for the same `groupId` subject (e.g. lines 44, 117, 209, 262, 315, 430, 494).

Could be a copy-paste from the `CREATE_GROUP` event in `createGroupController.ts` (where `NOT_APPLICABLE` is correct because there's no `groupId` yet). For the edit PDU handler, the group already exists and its ID is being passed as the subject.


```
> // Change:
> await sendAuditEvent('VIEW_EDIT_GROUP_PDU', username, groupId, 'NOT_APPLICABLE')
> // To:
> await sendAuditEvent('VIEW_EDIT_GROUP_PDU', username, groupId, 'SEARCH_TERM')
```

And update the matching test assertions in editGroupController.test.ts (search for 'VIEW_EDIT_GROUP_PDU') to expect 'SEARCH_TERM' instead of 'NOT_APPLICABLE'.

raby: Fair enough, thanks for the context. Firing before the call is the right call if that's the audit team's recommendation (captures intent regardless of downstream failure, and keeps a consistent pattern across the codebase).  Happy to mark this as resolved. 👍

raby (approved):
Looks great

PR #674, opened 2026-07-15
Apg 2053/remove additional width container v2

Alison-Weatherston:
Making the layout as consistent as possible
Fixing some accessibility best practise issues around landmarks
Fixing some small issue of missing quotes etc
Fixing the 'bounce' in sorting arrows


server/referralDetails/referralDetailsController.test.ts
```diff
@@ -55,6 +55,7 @@ describe('referral-details', () => {
           expect(res.text).toContain(referralDetails.crn)
           expect(res.text).toContain(referralDetails.personName)
           expect(res.text).toContain(personalDetails.probationDeliveryUnit)
+          expect(res.text).not.toContain('<div class="govuk-header">')
```
raby: `header.njk `uses `class="hmpps-header"`, not `govuk-header`, this assertion will always pass regardless of the change, so it's not testing anything useful. Did you mean `hmpps-header`? Just a check :-)

assets/scss/overrides/_local.scss
```diff
@@ -1,7 +1,4 @@
 @use '../../../node_modules/govuk-frontend/dist/govuk' as *;
-.govuk-main-wrapper {
```
raby, line 2: Worth a quick visual check on pages with minimal content (e.g. changeCohort, updateLdc), was this here  to stop the footer riding up?

server/referralDetails/referralDetailsController.test.ts
Alison-Weatherston: fixed - good spot.

assets/scss/overrides/_local.scss
Alison-Weatherston: If it comes up in design review, I will address it; otherwise i am happy to leave it as is.

raby (approved):

PR #676, opened 2026-07-15
added session tests

DanEgan1:


README.md
```diff
@@ -291,6 +291,24 @@ Or run tests with the cypress UI:
 
 `npm run int-test-ui`
 
+## Testing Audits locally 
+
+```sh
+
+# Start the services (the -d flag keeps your terminal clear)
+$ docker exec hmpps-mandd-localstack awslocal sqs create-queue --queue-name sqs-hmpps-audit
```
DanEgan1, line 299: Initial investigation suggests it may be possible to auto create the queue in the docker compose. Didn't do that in this PR to avoid doing too many things in one PR. Will investigate further and raise a separate change if proves a better solution.

In api its done as part of the build

server/testutils/factories/risksAndNeeds/groupSessionDetailsFactory.ts
```diff
@@ -1,9 +1,9 @@
 import { Factory } from 'fishery'
 import { GroupSessionResponse } from '@manage-and-deliver-api'
 
-class SessionDetailsFactory extends Factory<GroupSessionResponse> {}
+class GroupSessionDetailsFactory extends Factory<GroupSessionResponse> {}
 
-export default SessionDetailsFactory.define(() => ({
+export default GroupSessionDetailsFactory.define(() => ({
```
DanEgan1, line 6: Renamed factory to differentiate from new one

niaevans9 (dismissed):
lgtm

server/editSession/editSessionController.ts
```diff
@@ -431,6 +459,9 @@ export default class EditSessionController extends BaseController {
     let formError: FormValidationError | null = null
     let userInputData = null
 
+    if (req.method === 'GET') {
+      await sendAuditEvent('VIEW_EDIT_SESSION_FACILITATORS', username, groupId, 'SEARCH_TERM', { details: { groupId } })
```
raby: Every other session-level audit in this controller uses `sessionId` as the `subjectId`, is `groupId` intentional?

raby (dismissed):
Looks good. Just a small non blocking q

server/editSession/editSessionController.ts
DanEgan1: Good spot did a few of them the wrong way round initially (groupid as the subject session in the detail) Must have missed this when fixing them

raby (approved):

PR #679, opened 2026-07-16
Fix sentry implementation by ensuring its imported in app

Jack-Dempster:


DanEgan1 (approved):

raby (commented):
Looks good - could have optional sentryUser..test.ts
?

`it('returns consistent SHA-256 hash for a given username', () => {
  expect(hashUsername('testuser')).toBe(hashUsername('testuser'))
})
it('does not return the original username', () => {
  expect(hashUsername('testuser')).not.toContain('testuser')
})`

PR #683, opened 2026-07-17
Apg 2392/add audits to user actions missed audits

DanEgan1:


server/onboarding/onboardingController.ts
```diff
@@ -70,9 +75,6 @@ export default class OnboardingController {
         username: req.user.username,
         requestId: req.id,
         referralIdCount: referralIds.length,
-        successCount: response.successIds.length,
```
DanEgan1, line 73: The response from api changed to no longer include these after it was made fire and forget. It hadn't been updated here so it would error.

server/groupOverview/removeFromGroup/removeFromGroupController.ts
```diff
@@ -44,6 +44,13 @@ export default class RemoveFromGroupController extends BaseController {
       req.user.username,
     )
 
+    if (req.method === 'GET') {
```
DanEgan1, line 47: EDIT_REMOVE_FROM_GROUP was added in previous PR but view logs were missed

server/attendance/attendanceController.test.ts
```diff
@@ -943,5 +1072,12 @@ describe('showRecordAttendancePages', () => {
       .expect(res => {
         expect(res.text).toContain(`Select an attendance status for ${bffData.people[0].name}`)
       })
+    expect(sendAuditEvent).not.toHaveBeenCalledWith(
```
raby, line 1075: The expect block after return will never run. Jest reports the test as passing but the audit assertion is silently skipped.

Fix: Move the audit assertion inside the .expect() chain or use await + then:

```
it('Calling with incorrect parameters...', async () => {
  await request(app)
    .post(`/111/6789/record-attendance`)
    .type('form')
    .send({})
    .expect(400)
    .expect(res => {
      expect(res.text).toContain(`Select an attendance status for ${bffData.people[0].name}`)
    })

  expect(sendAuditEvent).not.toHaveBeenCalledWith(
    'VIEW_RECORD_ATTENDANCE',
    expect.anything(),
    expect.anything(),
    expect.anything(),
    expect.anything(),
  )
})
```

server/pni/pniController.ts
```diff
@@ -23,6 +24,9 @@ export default class PniController extends BaseController {
       referralId,
       username,
     )
+    await sendAuditEvent('VIEW_PNI', req.user.username, referralDetails?.crn, 'CRN', {
```
raby: Every other controller in this branch uses `if (req.method === 'GET') { ... }` before emitting a VIEW audit. If a POST route is ever added to this handler, the VIEW audit will fire on every method.

```
if (req.method === 'GET') {
  await sendAuditEvent('VIEW_PNI', req.user.username, referralDetails?.crn, 'CRN', {
    referralId,
  })
}
```

server/availabilityAndMotivation/locationPreferences/locationPreferencesController.ts
```diff
@@ -171,13 +171,13 @@ export default class LocationPreferencesController extends BaseController {
           data.paramsForUpdate.cannotAttendLocations
 
         await sendAuditEvent(
-          'UPDATE_REFERRAL_LOCATION_PREFERENCES',
+          'EDIT_REFERRAL_LOCATION_PREFERENCES',
```
raby, line 174: the `VIEW_CANNOT_ATTEND_LOCATIONS` call a few lines below is not guarded by `if (req.method === 'GET')`. Unlike `VIEW_ADD_LOCATION_PREFERENCES` and `VIEW_ADDITIONAL_PDU_LOCATION_PREFERENCES` elsewhere in this file which are both GET-guarded, this one sits after the POST block with no guard, so it fires on every POST that fails validation (falls through instead of returning early). 

It won't fire on a successful POST (that has an early return res.redirect), which is why the existing tests don't catch it. Fix is a one-liner, wrap it in `if (req.method === 'GET') `— and add a test for the POST + validation error case to lock it in.

Compare the two other methods in the same controller:

```
// showLocationPreferencesPage — CORRECT ✅
if (req.method === 'GET') {
  await sendAuditEvent('VIEW_ADD_LOCATION_PREFERENCES', ...)
}
// POST handling follows...

// showAdditionalPduLocationPreferencesPage — CORRECT ✅
if (req.method === 'GET') {
  await sendAuditEvent('VIEW_ADDITIONAL_PDU_LOCATION_PREFERENCES', ...)
}
// POST handling follows...

// showCannotAttendLocationsPage — BUG ❌
if (req.method === 'POST') {
  if (data.error) {
    formError = data.error   // no early return, falls through below
  } else {
    // ...early return on success
  }
}
// Fires for GET ✓ but ALSO fires for POST+error ✗
await sendAuditEvent('VIEW_CANNOT_ATTEND_LOCATIONS', ...)
```


Fix:


```
// showCannotAttendLocationsPage
if (req.method === 'POST') {
  const data = await new AddLocationPreferenceForm(req, referralId).cannotAttendLocationData()
  if (data.error) {
    res.status(400)
    formError = data.error
    userInputData = req.body
  } else {
    await sendAuditEvent('EDIT_REFERRAL_LOCATION_PREFERENCES', ...)
    // ...save + redirect
  }
}

// Guard now matches the other two methods in this file
if (req.method === 'GET') {
  await sendAuditEvent('VIEW_CANNOT_ATTEND_LOCATIONS', username, referralDetails?.crn ?? referralId, 'CRN', {
    referralId,
  })
}

// render page...
```

This bug is also undetected because `locationPreferencesController.test.ts` has no test for the POST + validation error path on `/add-locations-cannot-attend` , we could add one to validate.

raby (approved):
Looks good

PR #685, opened 2026-07-21
Apg 2433/upgrade add another

niaevans9:
Upgrading MOJ frontend to v10
New implementation of the add another component
Knock on upgrade of linting dependency so a handful of unrelated linting fixes 

image
image
image
image


raby (approved):
Great work I had a deep dive

PR #702, opened 2026-07-30
Fix group pages layout

niaevans9:
Before
Screenshot 2026-07-30 at 14 04 02
After
Screenshot 2026-07-30 at 14 03 48


Alison-Weatherston (approved):
LGTM

PR #709, opened 2026-08-03
Apg 1918/probation practiitioner null fix

Alison-Weatherston:
Where there is no practitioner name, 'No information available' will be displayed instead of 'null'.

Screenshot 2026-08-03 at 14 15 45


pfcairns (approved):
LGTM

raby (commented):
Looks good

PR #711, opened 2026-08-04
Apg 2103/capitalise m in midday and midnight

Alison-Weatherston:
**Midnight capitalised**

Screenshot 2026-08-04 at 14 18 30

**Midnight capitalised**

Screenshot 2026-08-04 at 14 18 40

**Midnight capitalised**

Screenshot 2026-08-04 at 14 18 51


raby (commented):
Every other presenter in this PR uses `timeWithCapitalisedMidday` directly, but `AttendanceHistoryPresenter` still manually capitalises session.time. The manual approach only capitalises the first character of the whole string. It works by accident for the factory's 'midday to 1pm' → 'Midday to 1pm' (midday happens to start the string), but silently fails for '9:30am to midday' → '9:30am to midday' (midday stays lowercase because it is mid-string)?

The fix is a one-liner:
```
// Before
{ text: session.time.charAt(0).toUpperCase() + session.time.slice(1) },

// After
{ text: session.timeWithCapitalisedMidday },
```

The factory already supplies the correct value:
```
// attendanceHistoryResponseFactory.ts
time: 'midday to 1pm',
timeWithCapitalisedMidday: 'Midday to 1pm',   // <-- this should be what's rendered
```

Sorry if I'm off and not needed there, ignore if so :-) 

That was my comment, an LLM came up with: (ignore if not relevent)

`formatStartDaysAndTimes` callers — confirm whether its output (which includes lowercase midday/midnight) is ever user-facing; capitalise if so.

server/sessionSchedule/sessionAttendance/sessionScheduleAttendancePresenter.test.ts
```diff
@@ -758,6 +764,7 @@ describe('SessionScheduleAttendancePresenter', () => {
                 participants: ['John Doe', 'Jane Smith', 'Bob Brown'],
                 dateOfSession: '1 June 2025',
                 timeOfSession: '10:00am to 12:00pm',
+                timeWithCapitalisedMidday: '10:00am to 12:00pm',
```
raby:
```
timeOfSession: '10:00am to 12:00pm',
timeWithCapitalisedMidday: '10:00am to 12:00pm',  // identical to timeOfSession
```

Both fields carry the same value, so the test would pass even if the presenter fell back to `timeOfSession` rather than `timeWithCapitalisedMidday`. Compare with the `groupSessionDetailsFactory` which deliberately differs:

```
time: '9:30am to midday',
timeWithCapitalisedMidday: '9:30am to Midday',   // ← values differ, test is meaningful
```

Suggest changing the test mock to:
```
timeOfSession: '10:00am to 12:00pm',
timeWithCapitalisedMidday: '10:00am to Midday',
```

raby (approved):
Looks great!

PR #721, opened 2026-08-06
APG-2472 Reapply LAO badge changes from PR #691

Alison-Weatherston:


server/config.ts
```diff
@@ -61,6 +61,7 @@ export default {
   enable_region_restriction: get('ENABLE_REGION_RESTRICTION', 'false', requiredInProduction) === 'true',
   enable_caselist_singular_result_text:
     get('ENABLE_CASELIST_SINGULAR_RESULT_TEXT', 'false', requiredInProduction) === 'true',
+  enable_restricted_access_badge: get('ENABLE_RESTRICTED_ACCESS_BADGE', 'true', requiredInProduction) === 'true',
```
raby: The fallback default is 'true', meaning any environment without this env var set (e.g. a new environment, a misconfigured deployment) will show the badge. Every other feature flag in this file defaults to 'false'. Given this surfaces LAO information, the default should also be 'false' to follow the pattern? (unless there is a reason why not?) Prod is explicitly set to "false" so is fine, but some defence applies here maybe? Just a thought

server/caselist/caseListUtils.ts
```diff
@@ -1,7 +1,14 @@
...
+import config from '../config'
 
 export default class CaselistUtils {
   static hasLdcTagHtml(referral: ReferralCaseListItem): string {
     return referral.hasLdc ? '</br><span class="moj-badge moj-badge--bright-purple">LDC</span>' : ''
   }
+
+  static hasLaoBadgeHtml(referral: ReferralCaseListItem): string {
+    return referral.lao && config.enable_restricted_access_badge
+      ? '</br><span class="moj-badge moj-badge--red">RESTRICTED ACCESS</span>'
```
raby: The </br> in `caseListUtils.ts `is non-standard (it's copied from the existing LDC badge pattern but should still be corrected). It works in most browsers via error recovery but fails strict HTML validation. Recommendation: Standardise to <br> to match `editSessionPresenter.ts` lines 163, 180? Small nit

server/editSession/editSessionPresenter.test.ts
```diff
@@ -497,6 +502,120 @@ describe('EditSessionPresenter', () => {
...
+              cells: [
+                { html: nameCrnCellHtml('456', 'Jane Doe', 'CRN002', 20, true) },
+                { html: '<span class="govuk-tag govuk-tag--red">Not attended</span>' },
+                {
+                  html: '<a href="/group-123/session-456/session-1-attendance-and-session-notes?referralId=456&source=edit-session">Jane Doe: Session 1 notes</a>',
+                },
+              ],
+            },
+          ],
+        })
```
raby, line 623: No tests for `enable_restricted_access_badge = false`
The tests in `editSessionPresenter.test.ts` and `groupAllocationsPresenter.test.ts` that cover LAO badge rendering rely on the config default being 'true' at test time. There is no test asserting the badge is absent when config.enable_restricted_access_badge is false (even when lao: true). 

We could add at least one test mocking `config.enable_restricted_access_badge = false` to verify the badge does not render:

Here's a sample test you can drop into editSessionPresenter.test.ts inside the existing describe('when a referral is an LAO case' block to cover the feature-flag-off scenario:
```

describe('when enable_restricted_access_badge feature flag is off', () => {
  beforeEach(() => {
    jest.resetModules()
    jest.mock('../../config', () => ({
      default: { enable_restricted_access_badge: false },
    }))
  })

  afterEach(() => {
    jest.resetModules()
  })

  it('does not render the RESTRICTED ACCESS badge even when referral is LAO', async () => {
    // Re-import presenter after mocking config
    const { default: EditSessionPresenterMocked } = await import('./editSessionPresenter')

    const sessionDetails: GroupSessionResponse = {
      pageTitle: 'Session 1',
      code: 'CODE-123',
      sessionType: 'Individual',
      isCatchup: false,
      attendanceAndSessionNotes: [
        {
          referralId: '123',
          name: 'Alex River',
          crn: 'CRN001',
          attendance: 'Attended',
          sessionNotes: 'Good progress',
          lao: true, // LAO is true, but flag is off
        },
      ],
      date: '01 Feb 2026',
      time: '1:00pm',
      timeWithCapitalisedMidday: '1:00pm',
      unformattedEndDate: '2026-02-01T14:00:00',
      scheduledToAttend: [],
      facilitators: [],
    }

    const presenter = new EditSessionPresenterMocked(mockGroupId, sessionDetails, mockSessionId, mockDeleteUrl)
    const result = presenter.attendanceTableArgs

    // Name cell should NOT contain the badge
    const rows = (result as { rows: { html: string }[][] }).rows
    expect(rows[0][0].html).not.toContain('RESTRICTED ACCESS')
    expect(rows[0][0].html).not.toContain('moj-badge--red')
  })
})
```

raby (approved):

PR #726, opened 2026-08-11
allow filtering by sex on the caselist

rgforsyth:
allow filtering by sex on the caselist

image


raby (approved):
LGTM - no blockers. Sex options are hardcoded for now which is fine given the API doesn't supply them.

PR #732, opened 2026-08-14
tidy up feature flag

Alison-Weatherston:


Jack-Dempster:
Should probably remove the values from the helm files as part of this but apart from that looks good!

helm_deploy/values-preprod.yaml
```diff
@@ -19,8 +19,6 @@ generic-service:
     SENTRY_ENVIRONMENT: preprod
     ENABLE_PAST_DATE_REGION_RESTRICTION: "true"
     ENABLE_REGION_RESTRICTION: "true"
-    ENABLE_CASELIST_SINGULAR_RESULT_TEXT: "true"
-    ENABLE_RESTRICTED_ACCESS_BADGE: "true"
```
Jack-Dempster, line 23: We might still need this one? Or if we are removing it from here then we will need to remove it from the code where its called

Alison-Weatherston: Do you mean this one? 

ENABLE_RESTRICTED_ACCESS_BADGE or 

ENABLE_CASELIST_SINGULAR_RESULT_TEXT

Both have had all references removed.

Jack-Dempster: ENABLE_RESTRICTED_ACCESS_BADGE still has references on the case list does it not? and all the pop pages?

Jack-Dempster (approved):
Looks good thanks for making those changes

